PR #18, opened 2017-12-11
Fix scalar types in oneof for proto3

amatalai:
### Example module:
```elixir
defmodule Oneof do
   use Protobuf, syntax: :proto3

   defstruct [:option]

   oneof :option, 0
   field :option_a, 1, optional: true, type: :string, oneof: 0
   field :option_b, 2, optional: true, type: :int32, oneof: 0
end
```

### Current behaviour (master branch):
```elixir
> Oneof.new(option: {:option_a, "string"}) |> Oneof.encode()
** (Protobuf.InvalidError) Oneof#option_a is invalid!
```

### Reason
Oneof fields are not present in generated struct and Map.get/1 in validator returns nil which is not allowed value for scalar types in proto3

### DIFF
--- a/lib/protobuf/validator.ex
+++ b/lib/protobuf/validator.ex
@@ -38,6 +38,8 @@ defmodule Protobuf.Validator do
       type_valid?(key_type, k) && match_and_valid?(val_type, v)
     end)
   end
+  # oneof fields are not present in struct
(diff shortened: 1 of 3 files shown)


tony612:
Thanks. It's a nice fix :)

PR #23, opened 2018-01-31
link to install readme and mention homebrew

Ch4s3:


tony612:
How about give this link https://github.com/google/protobuf#protocol-compiler-installation so that most people can choose the pre-built binary instead of building from source.

PR #25, opened 2018-02-14
Add Google Protobuf Timestamp type

zentetsukenz:
Add `Google.Protobuf.Timestamp` type. This allows protobuf file that contain `google.protobuf.Timestamp` type be able to use a generated code without any error.

Google Protobuf file for Timestamp type is also added as a reference for this new type.

Please note that, in proto3 version, variable will be marked as optional by default.

### DIFF
--- a/lib/google/protobuf/timestamp.pb.ex
+++ b/lib/google/protobuf/timestamp.pb.ex
@@ -1,0 +1,14 @@
+defmodule Google.Protobuf.Timestamp do
+  @moduledoc false
+  use Protobuf, syntax: :proto3
+
+  @type t :: %__MODULE__{
+          seconds: integer,
+          nanos: integer
+        }
+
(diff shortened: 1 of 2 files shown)


tony612:
Hey, I discuss this PR with some of other users and found a problem: Erlang/Elixir modules are global, so the same modules can't be defined more than once. If users want to use timestamp.pb.ex generated from a specific version protobuf(different from that in protobuf-elixir), then there'll be conflicts. So I'm considering move Google protobuf from protobuf-elixir and put them in a separated repo like https://github.com/tony612/google-protobuf. So developers can choose to add that package as dep. What's your thoughts?

zentetsukenz:
@tony612 Agreed.

By the way, I think it's probably be a good idea to let user be able to generate an Elixir module from protobuf file themselves.

PR #28, opened 2018-02-28
make google protobuf a standalone depend

falood:
this pr is depends on https://github.com/tony612/google-protobuf/pull/1
hold it until `google-protobuf` package released 🙂


mix.exs
```diff
@@ -57,7 +57,7 @@ defmodule Protobuf.Mixfile do
       maintainers: ["Tony Han"],
       licenses: ["MIT"],
       links: %{"GitHub" => "https://github.com/tony612/protobuf-elixir"},
-      files: ~w(mix.exs README.md lib config LICENSE priv/templates .formatter.exs)
+      files: ~w(mix.exs README.md lib/protobuf.ex lib/protobuf config LICENSE priv/templates .formatter.exs)
```
falood: @tony612 protos in `./lib/google` is only used by protoc plugin, so I think we don't need to package them in.

tony612: That's exactly what I want to do recently :)

PR #36, opened 2018-06-28
Add OTP 21.0 to build matrix, also try building a package

aquarhead:
Exclude 1.4 and 1.5 w/ OTP 21.0 since TravisCI doesn't support this (Hex does
not supply such combination yet?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,13 @@ elixir:
 otp_release:
   - 19.3
   - 20.0
+  - 21.0
+matrix:
+  exclude:
+  - elixir: 1.4
+    otp_release: 21.0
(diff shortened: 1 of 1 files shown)


amatalai:
> Exclude 1.4 and 1.5 w/ OTP 21.0 since TravisCI doesn't support this (Hex does
not supply such combination yet?)

1.6.6 is the earliest elixir version compatible with OTP21 https://hexdocs.pm/elixir/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp

aquarhead:
@amatalai thanks for the link :) was wondering where to find that information

PR #40, opened 2018-08-14
Fixed proto3 lost oneof values

### LINKED ISSUE #39: Oneof value is lost when set to type default value
**Protobuf:**
```proto
syntax = "proto3";

package example;

message Foo {
  oneof bar {
    int64 int = 1;
    string string = 2;
  }
}
```

**Example of invalid behavior:**
```elixir
iex(1)> Example.Foo.new(bar: {:string, "x"}) |> Example.Foo.encode |> Example.Foo.decode
%Example.Foo{bar: {:string, "x"}} # WORKING AS EXPECTED

iex(2)> Example.Foo.new(bar: {:string, ""}) |> Example.Foo.encode |> Example.Foo.decode
%Example.Foo{bar: nil} # VALUE LOST

iex(3)> Example.Foo.new(bar: {:int, 1}) |> Example.Foo.encode |> Example.Foo.decode
%Example.Foo{bar: {:int, 1}} # WORKING AS EXPECTED

iex(4)> Example.Foo.new(bar: {:int, 0}) |> Example.Foo.encode |> Example.Foo.decode     
%Example.Foo{bar: nil} # VALUE LOST
```

**Additional information:**
Issue was not present prior 4a2d61f050f322cfcbf1243f5d76e57085d56b49

amatalai:
fixes #39

### DIFF
--- a/lib/protobuf/encoder.ex
+++ b/lib/protobuf/encoder.ex
@@ -30,7 +30,7 @@ defmodule Protobuf.Encoder do
           syntax == :proto2 && ((val == nil && prop.optional?) || val == [] || val == %{}) ->
             acc
 
-          syntax == :proto3 && empty_val?(val) ->
+          syntax == :proto3 && ((empty_val?(val) && !oneof) || (oneof && is_nil(val))) ->
             acc
 
           true ->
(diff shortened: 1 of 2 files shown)


amatalai:
Hi @tony612 
is something blocking this from being merged?

PR #41, opened 2018-10-08
Generate typespecs for messages with no fields

JayH5:
Messages without fields still need typespecs, especially if those messages types are referenced from other message types.


skwerlman:
this pr is important, as the missing typespecs emit warnings in dialyxir

PR #43, opened 2018-12-26
supports atom value for string field

edwardzhou:
It is very convenience that supports atom value for string field

message Foo {
  string status = 1;
}

Foo.new(status: :ok)
|> Foo.encode()
|> Foo.encode()

==> %Foo{status: "ok"}


tony612:
I prefer not doing this. If you need string, then assign string.

PR #45, opened 2019-01-30
Add big int condition

leineu2016:
add a condition to handle big_unit and big_sint


leineu2016:
Close this pr, made by mistake.

PR #46, opened 2019-01-31
Add descriptors to generated protos

hassox:
There is a great deal of information in the descriptors that unfortunately is lost with the current approach.

Most protubuf libraries provide access to the raw descriptors, this PR brings this lib inline with that philosophy by
adding the encoded descriptor to the generated modules and exposing them via a `descriptor` function.

### DIFF
--- a/priv/templates/enum.ex.eex
+++ b/priv/templates/enum.ex.eex
@@ -4,6 +4,7 @@ defmodule <%= name %> do
 
   <%= if not is_nil(desc) do %>
   def descriptor do
+    # credo:disable-for-next-line
     <%= desc.__struct__ %>.decode(
       <%= desc.__struct__.encode(desc |> Map.from_struct() |> Enum.filter(fn {_, x} -> x != nil end) |> desc.__struct__.new()) |> inspect(limit: :infinity) %>
     )
--- a/priv/templates/enum.ex.eex
+++ b/priv/templates/enum.ex.eex
@@ -3,8 +3,10 @@ defmodule <%= name %> do
   use Protobuf<%= options %>
 
   def descriptor do
-    <%= desc.__struct__.encode(desc |> Map.from_struct() |> Enum.filter(fn {_, x} -> x != nil end) |> desc.__struct__.new()) |> inspect(limit: :infinity) %>
-    |> <%= desc.__struct__ %>.decode()
+    bin_desc = 
+      <%= desc.__struct__.encode(desc |> Map.from_struct() |> Enum.filter(fn {_, x} -> x != nil end) |> desc.__struct__.new()) |> inspect(limit: :infinity) %>
+
+    <%= desc.__struct__ %>.decode(bin_desc)
   end
 
 <%= Enum.map fields, fn(field) -> %>
--- a/priv/templates/enum.ex.eex
+++ b/priv/templates/enum.ex.eex
@@ -2,6 +2,11 @@ defmodule <%= name %> do
   @moduledoc false
   use Protobuf<%= options %>
 
+  def descriptor do
+    <%= desc.__struct__.encode(desc |> Map.from_struct() |> Enum.filter(fn {_, x} -> x != nil end) |> desc.__struct__.new()) |> inspect(limit: :infinity) %>
+    |> <%= desc.__struct__ %>.decode()
+  end
+
 <%= Enum.map fields, fn(field) -> %>
   field <%= field %>
 <% end %>
--- a/priv/templates/enum.ex.eex
+++ b/priv/templates/enum.ex.eex
@@ -3,10 +3,9 @@ defmodule <%= name %> do
   use Protobuf<%= options %>
 
   def descriptor do
-    bin_desc = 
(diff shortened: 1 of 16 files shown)


hassox:
Ping. Any movement here?

priv/templates/enum.ex.eex
```diff
@@ -2,6 +2,11 @@ defmodule <%= name %> do
   @moduledoc false
   use Protobuf<%= options %>
 
+  def descriptor do
```
KevinMarth: When the `Credo.Check.Readability.SinglePipe` is enabled in `credo`, each `descriptor` function causes the warning `[R] Use a function call when a pipeline is only one function long`. There are comment options to disable individual `credo` warnings, but these options work only on an entire source file or on a specified number of lines. The variable length of the byte stream to be decoded makes the latter option impractical.

I think it would be better to generate `descriptor` functions for `enum`, `message`, and `service` as follows.
```
def descriptor do
    Elixir.Google.Protobuf.DescriptorProto.decode(<<...>>)
end
```

KevinMarth:
This PR provides a needed public interface for type introspection via the `descriptor` functions. My project has previously been using the non-public and undocumented `__message_props__`, but this approach has the obvious downside.

priv/templates/enum.ex.eex
hassox: Updated to call the decode function with the binary directly

KevinMarth:
One other consideration is whether to update `tony612/google-protos` so that the `.pb.ex` files there also have `descriptor` functions.

priv/templates/enum.ex.eex
KevinMarth: One more `credo` issue. The call to the `decode` function causes the warning `[D] ↘ Nested modules could be aliased at the top of the invoking module.`

- An alias, e.g. `alias Google.Protobuf.DescriptorProto`, could be added to the modules and used in the call to the `decode` function.
- The `credo` comment `# credo:disable-for-next-line Credo.Check.Design.AliasUsage` could be added on the line above the call to the `decode` function.

tony612:
In fact, I don't really want to merge this PR because the generated code is redundant and I guess most people won't use `descriptor` function. If you want to get some metadata, you can use the `__message_props__` function. Could you explain your use cases?

hassox:
The __message_props__ loses a lot of information. Any custom options that are provided to a field or method are lost, even the json name option is not present. These extensions are often critical when building applications where options are used. Providing the descriptor in this way you get:

1: free support for any and all new options (including custom options)
2: a public API to get at this information described in protos

We've built heavily on this lib because we thought it is great but the lack access to the descriptor has made building tools with it very challenging. 

We've been unable to correctly use json names, access options related to Kafka event definitions, graphql annotations, http related options, permissions. There are many options to use (custom defined). If not by exposing the descriptor, how will this lib support accessing those?

connorjacobsen:
Protobuf libraries in almost every other language allow you to pull out custom options.
👍 to the Kafka & JSON name use cases.

hassox:
As an example of what bit me today... I want to use Jason to serialize these protocol buffer messages so I want to implement a protocol impl for each one. We use the standard `json_name` option in some places to specify what name we use but in most place we just want the normal camel casing. The json name option being unavailable means I can't fall back to the descriptor. Since this has lost fidelity in this way it is impossible for me to create tools based on this lib to generate compatible json representations.

If I had the descriptor available I wouldn't need to wait for a change to be made to this lib. I could get the information and move on.

connorjacobsen:
@tony612 Any updates here?

tony612:
@hassox Custom options is in my plan, but I need some time to finish it. For this PR, do you think using an protoc option can help you? Like
```
protoc --elixir_out=gen_ descriptor:. *.proto
```
You can refer to the current implementation https://github.com/tony612/protobuf-elixir/blob/173f6836dc840eee29f7acf62693403218913267/lib/protobuf/protoc/cli.ex#L24

So that people who don't need this won't be disturbed by the strange binary in generated code. Then I'll be very happy to merge this PR.

btw, I think some features like JSON converting can be supported by protobuf-elixir itself.

hassox:
The challenge is when we try to build tooling around our shared contracts we need options that are already accessible and familiar in the descriptor but will never end up in the main library. We use options for http, graphql and Kafka at the moment and likely will add more. 

Waiting for the library to explicitly handle an option just doesn't work out. If the descriptors were present, we could potentially contribute a bunch back to the library and help to flesh out elixir support for protocol buffer tooling built around the existing lib. That becomes difficult when there is a loss in consistency.

I don't mind making it an option though. How would you feel about making it opt out rather than opt in?

tony612:
@hassox The reason is I didn't see many requirements for this feature. So I'd rather making it opt out now instead of changing everyone's generated code by adding a strange function they don't know. I want to keep the generated code simple and easy to read.

hassox:
@tony612 I've updated to use a CLI option. Can you take a look mate?

tony612:
@hassox Done. Thanks!

PR #54, opened 2019-04-29
Add type_url/0 to Protobuf generated modules

matehat:
This is to support packing and unpacking `google.protobuf.Any` messages, as is being proposed in a separate PR in [tony612/google-protos](https://github.com/tony612/google-protos).

This compiles the type url as a static value exposed through a `type_url/0` functions in each module generated by the protobuf elixir compiler.

### DIFF
--- a/lib/protobuf.ex
+++ b/lib/protobuf.ex
@@ -1,5 +1,17 @@
 defmodule Protobuf do
   defmacro __using__(opts) do
+    path = __CALLER__.module
+      |> Module.split()
+      |> Enum.reverse()
+      |> Enum.with_index()
+      |> Enum.map(fn
+        {item, index} when index > 0 -> String.downcase(item)
+        {item, _} -> item
+      end)
+      |> Enum.reverse()
+      |> Enum.join(".")
+    type_url = "type.googleapis.com/" <> path
+
     quote do
       import Protobuf.DSL, only: [field: 3, field: 2, oneof: 2]
       Module.register_attribute(__MODULE__, :fields, accumulate: true)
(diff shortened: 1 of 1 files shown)


tony612:
I think a protobuf lib shouldn't handle this kind of special cases.

matehat:
Then, maybe just the "fully qualified name" attribute usually produced by protobuf libraries? The equivalent of https://github.com/protocolbuffers/protobuf/blob/b4f193788c9f0f05d7e0879ea96cd738630e5d51/python/google/protobuf/descriptor.py#L229-L231

matehat:
It's exactly equal to what follows `type.googleapis.com/` in a fully qualified message url

aquarhead:
I don't think `google.protobuf.Any` requires you to use that domain (`type.googleapis.com`) as a base for the type URL, us for example uses a custom one.

I basically forked this repo and added some additional support that suits our case in the `__using__` macro, link [here](https://github.com/ccpgames/protobuf-elixir/commit/8b7f7f513245c5d4fcbe184f91e8f8109ece92be). I don't think this PR should be merged in the current form unless we find a way to somehow configure this URL.

jdemilledt:
@matehat As mentioned in your other PR, I'd be willing to merge this into @demilletech's fork.

matehat:
Thanks @jdemilledt !

aquarhead:
Providing `type_fqdn` as `defoverridable` might not be the best way to customize it... Especially if generating code through the provided `protoc` plugin (I think?). You'd need to either manually add `def type_fqdn` to all generated modules or somehow modify the `protoc` plugin...

How about just putting it as an optional parameter, then user can find a performant way to feed that information in. I think [OTP 22's `persistent_term`](https://blog.erlang.org/persistent_term/) might be a good solution here? Or https://github.com/discordapp/fastglobal for older OTP releases.

jdemilledt:
@aquarhead @matehat Let's move this discussion to demilletech#3 since the maintainer here seems to dislike any functionality over the base.

PR #63, opened 2019-08-03
Add deprecated to field props

connorjacobsen:
Spent some time digging for the `deprecated` field option, realized it existed but was just excluded from the field options struct. Adding it here because we already have that data, and it's useful to be able to know programmatically if a field is deprecated.

Also updated the README with instructions on how to get the test suite running.


connorjacobsen:
Looks like the build failure was a network flake.

tony612:
Merged. Thank you!

PR #68, opened 2019-09-25
fix(encode): validate int range

### LINKED ISSUE #67: Value that is too big for an uint32 is successfully encoded and decoded
I have this protocol buffer definition:
```
syntax = "proto3";

message Uint {
  uint32 u = 1;
}
```
When I create a value that's too big for a 32 bit unsigned integer, encoding and decoding surprisingly works:
```
iex(10)> Uint.decode(Uint.encode(Uint.new(u: 12345678901234567890)))
%Uint{u: 12345678901234567890}
```
If I use a different library to decode the result, I get a shorter value:
```
iex(9)> Uint.encode(Uint.new(u: 12345678901234567890))
<<8, 210, 149, 252, 216, 206, 177, 170, 170, 171, 1>>
iex(10)> Proto2Test.Uint.decode(<<8, 210, 149, 252, 216, 206, 177, 170, 170, 171, 1>>)
%Proto2Test.Uint{u: 3944680146}
```
I think the `3944680146` value looks better, definitely fits into a 32 bit unsigned integer.

tony612:
Fixes #67

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,5 @@ script:
   - mix test
   - mix test.integration
   - mix hex.build
-  - mix format --check-formatted
+  # Only run for latest version
+  - '[[ "$TRAVIS_ELIXIR_VERSION" =~ 1.[6-8] ]] || mix format --check-formatted'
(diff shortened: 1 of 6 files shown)


whatyouhide:
Raising a `FunctionClauseError` is pretty hard to debug when this problem happens. Any chance that we can raise a more meaningful error?

tony612:
@whatyouhide Changed it a little bit. Thank you!

PR #70, opened 2019-11-05
Improve performance of __default_struct__/0

whatyouhide:
In this code we're building the default struct for Proto2 at compile time, and then building off from that struct to make the proto3 struct **at runtime**. There's no need to do that at runtime as far as I can tell, we can just do that at compile time too. That makes `MyProtoSchema.new()` about 3.5x faster from my benchmarks.

cc @tony612

### DIFF
--- a/lib/protobuf/dsl.ex
+++ b/lib/protobuf/dsl.ex
@@ -27,6 +27,11 @@ defmodule Protobuf.DSL do
     enum_fields = enum_fields(msg_props, false)
     default_struct = Map.put(default_fields, :__struct__, env.module)
 
+    default_struct_proto3 =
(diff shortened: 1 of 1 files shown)


relistan:
Would love to see this merged.

tony612:
Thanks! I don't know why I wrote code like that 🤦🏻‍♂️

relistan:
@tony612 Thank you! Can we get a release with this? 🚀

whatyouhide:
Ping @tony612, any chance to get a release for this? 🤗

tony612:
@whatyouhide I found an issue with this PR https://github.com/tony612/protobuf-elixir/issues/74

tony612:
@whatyouhide 0.7.0 is released with this.

PR #72, opened 2019-11-20
Adds a mix compiler task, based on `protoc`

jeanparpaillon:
This compiler can be added to project's list of compilers By default,
it will look for files in `proto` dir and output generated files in
`lib.proto`.

See `Mix.Tasks.Compile.Proto` module doc for additional options


tony612:
Sorry, I didn't see many benefits of having this. I prefer using `protoc`, which is simple, reliable and straightforward. But I guess you can make it a separated hex package shared within your company. Or you can use Makefile to avoid running `protoc` over and over.

PR #73, opened 2019-11-20
Encode required enums for proto2

mackeyja92:
This MR fixes the encoder when it is encoding required enums in proto2 by not skipping them when they are the "default" value.

A specific example that I have ran into that was causing issues is encoding `CommandAck` from the Pulsar protobuf [spec](https://github.com/apache/pulsar/blob/master/pulsar-common/src/main/proto/PulsarApi.proto). 

```protobuf
message CommandAck {
    enum AckType {
        Individual = 0;
        Cumulative = 1;
    }

    required uint64 consumer_id       = 1;
    required AckType ack_type         = 2;
    ...
}
```

```elixir
%Proto.CommandAck{
  ack_type: 0,
  consumer_id: 1,
  ...
}
```

`ack_type` is a required enum. This library was skipping encoding due to it being the default value which causes the pulsar server to close the connection and log an exception saying the message was missing fields. 

Encoding the above struct with this library outputs the following binary.

```elixir
<<8, 1, 26, 27, 8, 123, 16, 200, 3, 24, 255, 255, 255, 255, 255, 255, 255, 255, 255, 1, 32, 255, 255, 255, 255, 255, 255, 255, 255, 255, 1, 48, 0, 56, 0>>
```
I tested encoding the same struct with `:exprotobuf` library and the output is a little different.

```elixir
<<8, 1, 16, 0, 26, 27, 8, 123, 16, 200, 3, 24, 255, 255, 255, 255, 255, 255, 255, 255, 255, 1, 32, 255, 255, 255, 255, 255, 255, 255, 255, 255, 1, 48, 0, 56, 0>>
```

With this fix, the output from this library is the same as `:exprotobuf`.

### DIFF
--- a/lib/protobuf/encoder.ex
+++ b/lib/protobuf/encoder.ex
@@ -42,7 +42,7 @@ defmodule Protobuf.Encoder do
   end
 
   def encode_fields([prop | tail], syntax, struct, oneofs, acc) do
-    %{name_atom: name, oneof: oneof, enum?: is_enum, type: type} = prop
+    %{name_atom: name, oneof: oneof} = prop
 
     val =
       if oneof do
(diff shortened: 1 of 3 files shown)


tony612:
Thank you!

Issue #76, opened 2019-12-05
Raise exceptions if incorrect types are passed into `new`


keathley:
Currently, no checks are done when passing attributes to `new/2`. This means that it's possible to pass incorrect types or sub-messages while constructing a new struct. In most cases these errors won't be caught until the message is decoded since the encoder doesn't perform validations while it's encoding. To make this more concrete let's say that we have these definitions:

```elixir
defmodule Foo do
  @moduledoc false
  use Protobuf, syntax: :proto3

  defstruct [:id]

  field :id, 1, type: :int32
end

defmodule Bar do
  @moduledoc false
  use Protobuf, syntax: :proto3

  defstruct [:id]

  field :id, 1, type: :string
end

defmodule Baz do
  @moduledoc false
  use Protobuf, syntax: :proto3

  defstruct [:foo]

  field :foo, 1, type: Foo
end 
```

It's currently possible to do this:

```elixir
Baz.new(foo: Bar.new(id: "test")) |> Baz.encode()
<<10, 6, 10, 4, 116, 101, 115, 116>>
```

Errors like this won't be caught until the client attempts to decode the message on the other end.

## Proposal

I'd like to suggest that we check inputs as part of `new` and raise an exception if the value doesn't match the correct type. I believe that raising is the correct approach here because calling new with invalid data is equivalent to an `ArgumentError`. I don't believe that there's any recovery the user could take if they pass the wrong arguments to the function and its better to raise loudly so they're aware of the issue.

Obviously this would only work if the user calls `new` and doesn't work if they build the struct themselves. I'm not sure if that's an acceptable tradeoff or not.

I'm happy to work on submitting a PR for these changes if you believe that the idea has merit. But I wanted to make sure you thought this was a reasonable addition before I did the work on it.

whatyouhide:
> Obviously this would only work if the user calls new and doesn't work if they build the struct themselves. I'm not sure if that's an acceptable tradeoff or not.

This is how most things in Elixir work, so I strongly suggest that this is an acceptable tradeoff :)

@keathley we also don't check the field names in `new/1`, FWIW 😄

haljin:
I always wondered why aren't the defaults filled out when you just make the raw struct? You could feed default values to `defstruct` based off protobuf types?

PR #77, opened 2019-12-06
Raise on non-existing fields in Protobuf.Builder.new/2

whatyouhide:
cc @keathley as mentioned in #76 and cc @tony612 for review 🙃

### DIFF
--- a/lib/protobuf.ex
+++ b/lib/protobuf.ex
@@ -16,6 +16,10 @@ defmodule Protobuf do
       def new(attrs) do
         Protobuf.Builder.new(__MODULE__, attrs)
       end
+
+      def new!(attrs) do
+        Protobuf.Builder.new!(__MODULE__, attrs)
+      end
     end
   end
 
--- a/lib/protobuf/builder.ex
+++ b/lib/protobuf/builder.ex
@@ -4,37 +4,11 @@ defmodule Protobuf.Builder do
   end
 
   def new(mod, attrs) do
-    case attrs do
-      %{__struct__: _} ->
-        attrs
+    new_maybe_strict(mod, attrs, _strict? = false)
+  end
 
-      _ ->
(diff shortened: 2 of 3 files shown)


whatyouhide:
Ping @tony612 :)

tony612:
@whatyouhide Thanks for the PR. I'm thinking if it's better to have a `new!` function and keep the original one. The reason is sometimes I found it's useful to use a superset map to create a pb struct. How do you like it? @keathley

whatyouhide:
@tony612 we use this library pretty extensively at work and in my experience this is definitely an extremely unexpected behaviour, so much so that I've always considered this a bug. I think it's much more common to misspell a field name and only catch it when encoding than to desire this behaviour. Ultimately, up to you, I can create new! :)

tony612:
@whatyouhide I guess that's why we have both `struct` and `struct!`? Sometimes we specify the keys explicitly
```
Foo.new(a: 1, b: "foo")
```
But sometimes we use a map to build it like
```
map = got_from_somewhere_else()
Foo.new(map)
```
It's hard to guarantee the `map` in the later example match very well with our struct, especially in a dynamic language like Elixir.

whatyouhide:
@tony612 okay, we'll switch completely to `new!` and forget about `new` then 😄 I updated the PR, can you take a look?

tony612:
@whatyouhide Looks like `new!` should be used in the test?

whatyouhide:
Yep, I actually forgot to push this code but we need to generate the `new!/1` function in protobuf modules too 😄 I just pushed the new code and the updated test!

keathley:
Sorry I didn't respond earlier. I'm glad y'all ended up using `new!` since it allows `new` to remain "open" which should allow more re-use of existing maps. :+1:

PR #80, opened 2019-12-19
Generate enum types

haljin:
The protoc plugin generates structs for `message`s together with their types. It doesn't however do that for `enum`s and it could be very helpful to generate types for enums too, so that they can be reffered to in other areas of the code.

Let me know what you think :)

### DIFF
--- a/lib/protobuf/protoc/generator/enum.ex
+++ b/lib/protobuf/protoc/generator/enum.ex
@@ -10,8 +10,18 @@ defmodule Protobuf.Protoc.Generator.Enum do
     fields = Enum.map(desc.value, fn f -> generate_field(f) end)
     msg_name = Util.mod_name(ctx, ns ++ [name])
     generate_desc = if ctx.gen_descriptors?, do: desc, else: nil
+    type = generate_type(desc.value)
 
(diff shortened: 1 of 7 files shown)


lib/protobuf/protoc/generator/enum.ex
```diff
@@ -10,8 +10,18 @@ defmodule Protobuf.Protoc.Generator.Enum do
...
+    type = generate_type(desc.value)
 
-    Protobuf.Protoc.Template.enum(msg_name, msg_opts(ctx, desc), fields, generate_desc)
+    Protobuf.Protoc.Template.enum(msg_name, msg_opts(ctx, desc), fields, type, generate_desc)
+  end
+
+  def generate_type(fields) do
+    field_values =
+      fields
+      |> Enum.map(fn f -> ":#{f.name} | #{f.number}" end)
```
tony612: I think we only need `integer` instead of every number because the reason we may get integer as enum is the protobuf may include an unknown enum, which should also be handled based on Protobuf's spec.

haljin: Sure sounds reasonable :)

amiel:
I am looking forward to being able to use this. Has it been released?

tony612:
@amiel No. I will release the new version recently with other changes.

amiel:
Thanks!

PR #81, opened 2019-12-19
Raise DecodeError if the binary cannot be parsed

### LINKED ISSUE #71: Function clause exception when decoding garbage binaries
Hi, 

When I try to decode an obviously wrong binary as a protobuf struct I get a function clause exception, e.g.:

```
iex(2)> Protobuf.Decoder.decode(<<1,1,1>>, SomeProtoStruct)
** (FunctionClauseError) no function clause matching in Protobuf.Decoder.raw_decode_value/3    
    
    The following arguments were given to Protobuf.Decoder.raw_decode_value/3:
    
        # 1
        1
    
        # 2
        <<1, 1>>
    
        # 3
        [1, 0]
    
    Attempted function clauses (showing 4 out of 4):
    
        def raw_decode_value(0, <<bin::bitstring()>>, result)
        def raw_decode_value(2, <<bin::bitstring()>>, result)
        def raw_decode_value(5, <<n::integer()-size(32), rest::bitstring()>>, result)
        def raw_decode_value(1, <<n::integer()-size(64), rest::bitstring()>>, result)
    
    (protobuf) lib/protobuf/decoder.ex:283: Protobuf.Decoder.raw_decode_value/3
    (protobuf) lib/protobuf/decoder.ex:13: Protobuf.Decoder.decode/2
```

The thing is, I'd like to be able to distinguish these errors from other function clause exceptions that might happen in the same process. Currently the only way is to wrap the decode like this:

```
 defp decode(msg) do
    SomeProtoStruct.decode(msg)
  catch
    _, _ ->
      raise Protobuf.DecodeError
  end
```

Could the auto-generated decode function not do this? I'm willing to make a PR for this if you agree.

haljin:
Closes #71

### DIFF
--- a/lib/protobuf/decoder.ex
+++ b/lib/protobuf/decoder.ex
@@ -14,6 +14,9 @@ defmodule Protobuf.Decoder do
     %{repeated_fields: repeated_fields} = msg_props = module.__message_props__()
     struct = build_struct(kvs, msg_props, module.new())
     reverse_repeated(struct, repeated_fields)
+  catch
+    :error, :function_clause ->
+      raise Protobuf.DecodeError, message: "cannot decode binary data as #{module}"
   end
 
   @doc false
--- a/lib/protobuf/decoder.ex
+++ b/lib/protobuf/decoder.ex
@@ -14,9 +14,6 @@ defmodule Protobuf.Decoder do
     %{repeated_fields: repeated_fields} = msg_props = module.__message_props__()
     struct = build_struct(kvs, msg_props, module.new())
     reverse_repeated(struct, repeated_fields)
-  catch
-    :error, :function_clause ->
-      raise Protobuf.DecodeError, message: "cannot decode binary data as #{module}"
   end
 
   @doc false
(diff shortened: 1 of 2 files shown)


lib/protobuf/decoder.ex
```diff
@@ -14,6 +14,9 @@ defmodule Protobuf.Decoder do
     %{repeated_fields: repeated_fields} = msg_props = module.__message_props__()
     struct = build_struct(kvs, msg_props, module.new())
     reverse_repeated(struct, repeated_fields)
+  catch
+    :error, :function_clause ->
+      raise Protobuf.DecodeError, message: "cannot decode binary data as #{module}"
```
tony612: If we just want to catch this kind of errors like your example, I prefer adding a catching-all pattern match for `raw_decode_value`.

haljin: In this case the `function_clause` happens in `raw_decode_varint` though - should I instrument both functions?

tony612: OK.

Issue #89, opened 2020-03-26
Elixir terms converter extensions


ulissesalmeida:
Hi 👋 

I would like to know if the maintainers are interested in a feature like this.

Let's say I have this message:

```proto
package MyApp.Proto;

message NaiveDateTime {
  uint32 microseconds = 1;
}
```

Then, when I working I'm my app, I've always to convert to an Elixir term.

```elixir
message.birthday
%MyApp.Proto.NaiveDateTime{microseconds: ...}

# oh no, I have to remember to:
to_naive_date_time(message.birthday)
~N[1988-10-29 00:00:00]
```

I was wondering, wouldn't it be nice to have a way to tell the encoder/decoder how they work with some messages. Then, I could write in my app:

```elixir
message.birthday
~N[2000-01-01 00:00:00]
```

I really would like to help to add this feature, but I would like to hear from you folks if you are interested.

Then, we can discuss the API and how this extension should work.

I had a suggestion, we could have something like:

```proto
message NaiveDateTime {
  option (elixirpb.message).converter = "MyApp.NaiveDateTimeConverter";
  uint32 microseconds = 1;
}
```

Then in my app, I implement some module behavior:

```elixir
defmodule MyApp.NaiveDateTimeConverter do
  @behaviour Protobuf.Converter

  # or post_decode, decode, after_decode
  def to_elixir(%MyApp.Protobuff.NaiveDateTime{microseconds: microseconds}),
    do: NaiveDateTime.from_unix!(microseconds, :microsecond)

  # or pre_encode, encode, before_encode
  def to_protobuf(%DateTime{} = datetime) do
    microseconds = DateTime.to_unix(datetime, :microsecond)
    %MyApp.Protobuff.NaiveDateTime{microseconds: microseconds}
  end

  def to_protobuf(%NaiveDateTime{} = naive) do
    {:ok, datetime} = DateTime.from_naive(naive, "Etc/UTC")
    microseconds = DateTime.to_unix(datetime, :microsecond)
    %MyApp.Protobuff.NaiveDateTime{microseconds: microseconds}
  end

  def elixir_type, do: "NaiveDateTime.t() | DateTime.t()"
end
```

What do you folks think? Do we have space in this library for such a feature?

ulissesalmeida:
The other idea for API is would be using protocols. 
Something like:

```elixir
message NaiveDateTime {
  option (elixirpb.message).elixir_type = "NaiveDateTime.t() | DateTime.t()";
[...]

PR #91, opened 2020-03-27
Add Encodable/Decodable protocols

ulissesalmeida:
This feature adds the ability to convert to and from protobufs structs and elixir terms.

The ideas are from this issue: #89 

Usage:

If you want to transform an automatically a protobuf message to an Elixir term after decoding, you can implement `to_elixir` function of the `Protobuf.Decodable`

```elixir
      defimpl Protobuf.Decodable, for: MyApp.Protobuf.Date do
        def to_elixir(%MyApp.Protobuf.Date{year: year, month: month, day: day}) do
          {:ok, date} = Date.new(year, month, day)
          date
        end
      end
      # later in a decoded message
      proto_message.birthday
      ~D[1988-10-29]
```

If you want to automatically encode a elixir term to a protobuf message, you can implement the Encodable protocol.

```elixir
      defimpl Protobuf.Encodable, for: Date do
        def to_protobuf(%Date{year: year, month: month, day: day}, MyApp.Protobuf.Date) do
          MyApp.Protobuf.Date.new(year: year, month: month, day: day)
        end
      end
      # later, you can use Elixir terms in your fields and those will be
      # converted to protobuf structs before binary encoding
      %{protobuf_message | birthday: ~D[1988-10-29]}
```

Then, if you do this, basically you need to annotate your types differently, then you can annotate your protobuf message with the Elixir message:

```
       message Date {
         option (elixirpb.message).typespec = "Date.t";
         int iso_days = 1;
       }
```

### DIFF
--- a/test/protobuf/protoc/integration_test.exs
+++ b/test/protobuf/protoc/integration_test.exs
@@ -55,4 +55,48 @@ defmodule Protobuf.Protoc.IntegrationTest do
   test "extensions" do
     assert "hello" == Protobuf.Protoc.ExtTest.Foo.new(a: "hello").a
   end
+
+  describe "message extensions" do
+    setup do
+      %{protobuf_file: File.read!("test/protobuf/protoc/proto_gen/extension.pb.ex")}
+    end
+
+    test "uses the option type_name", %{protobuf_file: protobuf_file} do
+      assert protobuf_file =~ "inserted_at: DateTime.t() | nil\n"
+    end
+
+    test "keeps the same module", %{protobuf_file: protobuf_file} do
+      assert protobuf_file =~ "type: Protobuf.Protoc.ExtTest.UnixDateTime\n"
+    end
+  end
+
+  describe "encodable protocol" do
+    test "transform to protobuf before" do
+      msg =
+        Protobuf.Protoc.ExtTest.FooWithUnixDateTime.new(
+          inserted_at: udatetime(~N[2020-04-03 11:28:21.929987])
+        )
+
+      assert Protobuf.encode(msg) ==
+               <<10, 9, 8, 131, 160, 134, 184, 147, 204, 232, 2>>
+    end
+  end
+
+  describe "decodable protocol" do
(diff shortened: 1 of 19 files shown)


test/protobuf/protoc/integration_test.exs
```diff
@@ -55,4 +55,12 @@ defmodule Protobuf.Protoc.IntegrationTest do
   test "extensions" do
     assert "hello" == Protobuf.Protoc.ExtTest.Foo.new(a: "hello").a
   end
+
+  describe "message extensions" do
+    test "uses the option type_name" do
+      file = File.read!("test/protobuf/protoc/proto_gen/extension.pb.ex")
+
+      assert file =~ "inserted_at: DateTime.t() | nil\n"
```
ulissesalmeida: Very weak test, but I don't how is the best way to test generation from protobuf file here.

test/protobuf/protoc/proto_gen/extension.pb.ex
```diff
@@ -9,3 +9,27 @@ defmodule Protobuf.Protoc.ExtTest.Foo do
...
+defmodule Protobuf.Protoc.ExtTest.FooWithDateTime do
+  @moduledoc false
+  use Protobuf, syntax: :proto2
+
+  @type t :: %__MODULE__{
+          inserted_at: DateTime.t() | nil
+        }
+  defstruct [:inserted_at]
+
+  field :inserted_at, 1, optional: true, type: DateTime
```
ulissesalmeida, line 34: Oops, this is a bug, the `type` here should point to `Protobuf.Protoc.ExtTest.UnixDateTime`

ulissesalmeida: Fixed

whatyouhide:
Hey @ulissesalmeida, we decided that we can't take any new code in this library for now since we're trying to get it back in shape and releasing it. We want to keep only the features that we have right now without adding anything for the time being. We'll consider something like this before 1.0 though! Thank you so much 💟

PR #99, opened 2020-04-18
Add nested messages parse.

csWen:
The changes are in two files:
- /frame/metadata: implemented needed functions for `protoc-gen-validate`, not all functions
- /frame/parser.ex: implemented parsing nested messages

### DIFF
--- a/lib/protobuf/protoc/frame/parser.ex
+++ b/lib/protobuf/protoc/frame/parser.ex
@@ -44,7 +44,7 @@ defmodule Protobuf.Protoc.Parser do
        ) do
     ctx = %{ctx | package: package, syntax: syntax(syntax)}
 
-    msgs = Enum.map(desc.message_type, &parse_msg(ctx, package, nil, &1))
+    msgs = Enum.map(desc.message_type, &parse_msg(ctx, package, &1))
     # result = Result.put_tree([package, name], file)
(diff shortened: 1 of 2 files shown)


lib/protobuf/protoc/frame/parser.ex
```diff
@@ -65,23 +65,29 @@ defmodule Protobuf.Protoc.Parser do
     file
   end
 
-  @spec parse_msg(Context.t(), String.t(), Google.Protobuf.DescriptorProto.t()) :: any
-  defp parse_msg(%{namespace: ns} = ctx, file_fqn, desc) do
-    fqn = Utils.join_name([file_fqn, desc.name])
+  @spec parse_msg(Context.t(), String.t(), String.t(), Google.Protobuf.DescriptorProto.t()) :: any
+  defp parse_msg(_ctx, _file_fqn, _parent_msg, []), do: []
+
+  defp parse_msg(%{namespace: ns} = ctx, file_fqn, parent_msg_fqn, desc) do
```
tony612, line 71: I think we can change `file_fqn, parent_msg_fqn` to a single `parent_fqn`?

csWen: I thought we can get all messages(include nested messages) in a file if we have the file_fqn in message metadata. I can remove it if it's not useful.

PR #101, opened 2020-04-28
Option to return maps with :__msg__ key instead of structs

vans163:
A little hacky but I could not find a better way to do this without breaking some stuff. 

I am open to clean it up based on suggestions, but the basic idea is to have a way to return maps insteads of structs.  Structs are not very flexible (example put_in does not work) and other shenanigans.


whatyouhide:
@vans163 you can always convert a struct to a map yourself via `Map.from_struct/1`. Closing this as we don't plan to support it in the library itself. Thanks anyways! 💟

vans163:
> @vans163 you can always convert a struct to a map yourself via `Map.from_struct/1`. Closing this as we don't plan to support it in the library itself. Thanks anyways! heart_decoration

You can, then you will have the outer struct as a map but all nested structs + lists of structs will still be structs.

Np.

Issue #86, opened 2020-03-18
Feature Request: Encode to / decode from JSON


jahell:
In proto3 google added an official JSON mapping (https://developers.google.com/protocol-buffers/docs/proto3#json). Is there any plan to support encoding to and decoding from JSON?

whatyouhide:
I would agree to add this feature to this library. I would probably go with a `to_json/1` function and a `from_json/1` function in the generated Protobuf modules, and I would go with a customizable JSON encoder/decoder (so you can choose Jason, Poison, and so on).

tony612:
I think it's a good idea to support this, but there's no plan for the moment.

amiel:
Hi, I'm starting to work on implementing this.

@tony612 how do you feel about depending on https://github.com/michalmuskala/jason/? Any other thoughts you might have for me?

whatyouhide:
Hey @amiel! The company I work at ([Community](https://www.community.com)) is sponsoring some work on this library and @britto already started work on this. :) Feel free to coordinate with him here if you want to!

amiel:
@whatyouhide Awesome!

@britto I have started an implementation in my application (not in a fork of protobuf-elixir) and I'd be happy to share what I have so far.

britto:
Hi @amiel, that's awesome! Yes, I'd love to iterate on it with you.

amiel:
@britto Awesome. Feel free to send me an email at amiel@actionsprout.com.

whatyouhide:
Hey everyone, this has been implemented for a while. It's not in a main release yet but we'll do our best to release a new version soon. Anyways, closing this issue since the feature's on the main branch 😌

PR #105, opened 2020-06-12
New feature: JSON encoding

britto:
Addresses the first part of #86.

This feature allows us to encode messages as JSON in addition to protobuf binaries.

It follows Google's specs and reference implementations. Only `proto3` syntax is supported at the moment and some features, such as well-known types, are not fully supported yet. JSON encoding relies on the `Jason` library. Support to other JSON engines may be added in the future.

https://developers.google.com/protocol-buffers/docs/reference/google.protobuf

Huge thanks to @whatyouhide and @amiel for their support and effort put into this.

### TODO

- [x] Find more edge cases to cover and test
- [x] Document functions and their options
- [x] Compare JSON output against reference implementations (Go, Python, Java, C++)

### Out of scope

- Define `from_json` and `to_json` helpers for each generated module.
- `proto2` support
- `Any` mapping
- Google's well-known types mapping
- Runtime type checking


lib/protobuf/json.ex
```diff
@@ -0,0 +1,99 @@
...
+    end
+  end
+
+  defp get_props(%module{}) do
+    case module.__message_props__() do
+      %{syntax: :proto3} = props ->
+        {:ok, props}
+
+      %{syntax: syntax} ->
+        {:error, EncodeError.new({:unsupported_syntax, syntax})}
```
britto, line 86: Not sure if it is possible to embed proto2 messages into proto3 and vice versa. If it is, we should probably move this validation in, so we check it recursively too.

lib/protobuf/json.ex
```diff
@@ -0,0 +1,99 @@
...
+      end
+
+  With `encode/1` you can turn any `Protobuf` message struct into a JSON string:
+
+      iex> message = Car.new(color: :RED, top_speed: 125.3)
+      iex> Protobuf.JSON.encode(message)
+      {:ok, "{\\"color\\":\\"RED\\",\\"topSpeed\\":125.3}"}
+
+  JSON keys are encoded as camelCase strings, specified by the `json_name` field options. So
+  make sure to recompile the `.proto` files in your project before working with JSON encoding.
```
britto, line 54: We should probably mention this on the release notes.

jahell:
I wrote a version of this for our use case at Brex. Happy to help out with some reviews or share the code for that if that's useful to you.

>  Find more edge cases to cover and test
The test cases for the python implementation are pretty good. I just looked that those and basically copied the behavior.

lib/protobuf/json/encode.ex
```diff
@@ -0,0 +1,129 @@
...
+
+    for {oneof_name, _index} <- oneofs,
+        tag_and_value = Map.get(struct, oneof_name) do
+      {tag, value} = tag_and_value
+      prop = field_props[field_tags[tag]]
+      encode_field(prop, value, opts)
+    end
+  end
+
+  # TODO: handle invalid values? check types?
```
britto, line 44: When structs are built straight out of a binary using `Protobuf.decode` we know for sure they are well-formed and will encode to JSON correctly. Manipulated structs however may crash this at some point. I wonder how strict we should be in this regard without compromising performance.

britto:
Thanks, @jahell! Does this current feature set cover you use case? I'd love to test this out in a production-like environment and make all the needed adjustments soon.

I've been looking mostly into Golang's test cases, but the Python ones are definitely more readable. I'll try my best to incorporate them both here.

lib/protobuf/json.ex
```diff
@@ -0,0 +1,161 @@
+defmodule Protobuf.JSON do
+  @moduledoc """
+  JSON encoding and decoding utilities for Protobuf structs.
+
+  It follows Google's [specs](https://developers.google.com/protocol-buffers/docs/proto3#json)
+  and reference implementation. Only `proto3` syntax is supported at the moment. Some features
+  as [well-known](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf)
```
whatyouhide, line 7:
```suggestion
  such as [well-known](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf)
```

lib/protobuf/json.ex
whatyouhide: Should we also mention that the default is the camelcased field name? It's not too clear from the description above :)

lib/protobuf/json.ex
```diff
@@ -0,0 +1,161 @@
...
+  """
+
+  alias Protobuf.JSON.{Encode, EncodeError}
+
+  @type encode_opt :: {:json_library, module} | Encode.encode_opt()
+
+  @doc """
+  Generates a JSON representation of the given protobuf `struct`.
+
+  Similar to `encode/1` except it will unwrap the error tuple and raise in case of errors.
```
whatyouhide, line 74:
```suggestion
  Similar to `encode/2` except it will unwrap the error tuple and raise in case of errors.
  Accepts the same options as `encode/2`.
```

lib/protobuf/json.ex
```diff
@@ -0,0 +1,161 @@
...
+        use Protobuf, syntax: :proto3, enum: true
+        field :GREEN, 0
+        field :RED, 1
+      end
+
+      defmodule Car do
+        use Protobuf, syntax: :proto3
+        field :color, 1, type: Color, enum: true
+        field :top_speed, 2, type: :float, json_name: "topSpeed"
+      end
```
whatyouhide, line 118: This is a minor nitpick but I _think_ it would be nicer to show these as Protobuf schemas, and not as their compiled Elixir output (since you usually never touch that by hand)?

lib/protobuf/json.ex
```diff
@@ -0,0 +1,161 @@
...
+
+      %{syntax: syntax} ->
+        {:error, EncodeError.new({:unsupported_syntax, syntax})}
+    end
+  end
+
+  defp get_json_library(opts) do
+    json_library = Keyword.get(opts, :json_library, Jason)
+
+    if Code.ensure_loaded?(json_library) do
```
whatyouhide, line 155: I think that if we want to make the JSON library customizable, we should make sure it exports the right functions. Personally, I'd force folks to use Jason for starters and then wait for complaints if someone is not okay with this 😄 

In case we want to keep it customizable:

```suggestion
    if Code.ensure_loaded?(json_library) and function_exported?(json_library, :encode, 1) do
```

In case we want to just check that Jason is available, I'd go with

```elixir
if Code.ensure_loaded?(Jason) do
  ...
```

lib/protobuf/json/encode.ex
```diff
@@ -0,0 +1,139 @@
+defmodule Protobuf.JSON.Encode do
+  @moduledoc """
+  Internal JSON encoding API.
+  """
```
whatyouhide, line 4: Let's make this module `@moduledoc false` and document the option types in the `Protobuf.JSON` module.

lib/protobuf/json.ex
```diff
@@ -0,0 +1,161 @@
...
+      |> json_library.encode()
+    end
+  end
+
+  defp get_props(module) do
+    case module.__message_props__() do
+      %{syntax: :proto3} = props ->
+        {:ok, props}
+
+      %{syntax: syntax} ->
```
whatyouhide, line 147: I like me some extra safety :)

```suggestion
      %Protobuf.MessageProps{syntax: :proto3} = props ->
        {:ok, props}

      %Protobuf.MessageProps{syntax: syntax} ->
```

lib/protobuf/json/encode.ex
```diff
@@ -0,0 +1,139 @@
...
+    end
+  end
+
+  # TODO: handle invalid values? check types?
+  defp encode_field(prop, value, opts) do
+    {encode_key(prop, opts), encode_value(value, prop, opts)}
+  end
+
+  defp encode_key(prop, opts) do
+    if opts[:use_proto_names], do: prop.name, else: prop.json_name
```
whatyouhide, line 55: We default `prop.json_name` to the camelCased version of the Protobuf name if someone doesn't specify `json_name = ...`, right?

lib/protobuf/json/encode.ex
```diff
@@ -0,0 +1,139 @@
...
+  defp encode_value(value, %{type: type} = prop, _opts) when type in @raw_types do
+    maybe_repeat(prop, value, & &1)
+  end
+
+  defp encode_value(value, %{type: type} = prop, _opts) when type in @int64_types do
+    maybe_repeat(prop, value, &Integer.to_string/1)
+  end
+
+  defp encode_value(value, %{type: :bytes} = prop, _opts) do
+    maybe_repeat(prop, value, &Base.encode64/1)
```
whatyouhide, line 74: Do we care in any way (in JSON, in general) if this base64 is lowercase or uppercase?

lib/protobuf/json/encode.ex
```diff
@@ -0,0 +1,139 @@
...
+    %{field_props: field_props, field_tags: field_tags} = module.__message_props__()
+    key_prop = field_props[field_tags[:key]]
+    value_prop = field_props[field_tags[:value]]
+
+    for {key, val} <- map, into: %{} do
+      name = encode_value(key, key_prop, opts)
+      value = encode_value(val, value_prop, opts)
+
+      {to_string(name), value}
+    end
```
whatyouhide, line 101: For performance since this is optimized code:

```suggestion
    for {key, val} <- map, into: %{} do
      name = encode_value(key, key_prop, opts)
      value = encode_value(val, value_prop, opts)

      {to_string(name), value}
    end
```

lib/protobuf/protoc/generator/message.ex
```diff
@@ -287,9 +291,21 @@ defmodule Protobuf.Protoc.Generator.Message do
...
     |> Map.put(:packed, f.options.packed)
     |> Map.put(:deprecated, f.options.deprecated)
   end
+
+  # Omit `json_name` from the options list when it matches the original field
+  # name to keep the list small. Only Proto3 has JSON support for now.
+  defp put_json_name(opts, :proto3, %{name: name, json_name: name}), do: opts
+
+  defp put_json_name(opts, :proto3, %{json_name: json_name}) do
+    Map.put(opts, :json_name, inspect(json_name))
```
whatyouhide, line 307: Why are we `inspect/1`ing here? What do we get as `json_name`?

lib/protobuf/protoc/generator/message.ex
```diff
@@ -238,9 +238,13 @@ defmodule Protobuf.Protoc.Generator.Message do
...
   end
 
-  defp field_options(f) do
-    opts = %{enum: f.type == :TYPE_ENUM, default: default_value(f.type, f.default_value)}
-    if f.options, do: merge_field_options(opts, f), else: opts
+  defp field_options(f, syntax) do
+    enum = f.type == :TYPE_ENUM
+    default = default_value(f.type, f.default_value)
+
+    %{enum: enum, default: default}
```
whatyouhide, line 245: Nitpick :)
```suggestion
    enum? = f.type == :TYPE_ENUM
    default = default_value(f.type, f.default_value)

    %{enum: enum?, default: default}
```

lib/protobuf/json.ex
britto: Excellent idea! 💡

lib/protobuf/json.ex
britto: Good point. I think this was a leftover from previous iterations, when I was afraid I had to inject that dependency in the tests. Turns out I didn't. I'm going with the latter approach and keep it hardcoded for now.

lib/protobuf/json/encode.ex
britto: Yes, we use whatever `protoc` generates for us in that case.

lib/protobuf/json/encode.ex
britto: Good question. I'll double-check that.

lib/protobuf/protoc/generator/message.ex
britto: It's a string, we call `inspect` here in order to quote it correctly in the Elixir output, otherwise it will come out as a variable. There are some tests in place to avoid regressions if we eventually try to refactor this.

lib/protobuf/json/encode.ex
britto: Base64 itself is case-sensitive, so it is important to respect that. Map keys can be `bytes`, but they will always be encoded/decoded according to the rules for `bytes`, so it won't be a problem.

whatyouhide: Great 👍

lib/protobuf/json/encode.ex
whatyouhide: Ah, nice, good to know.

lib/protobuf/json/encode.ex
britto: I think we can delve deeper into this on a separate Issue/PR. As far as my research goes, all the dynamic languages implementations do type checking at the moment the fields are set, which seems like a good idea because if something is wrong with our data we get early feedback by failing fast, as opposed to doing it at encoding/serialization time. In the end I think the challenge will be finding a sensible way of doing that without compromising  too much the niceties of working with structs naturally.

amiel:
Hi @britto .

I haven't done any manual testing with this yet (I will, but it might be a couple of days before I can dive in to that).

One thing I think I'll need is a way to get the map representation (the result _before_ calling `json_library.encode()`). It looks like the only way to do this at the moment is to call `Protobuf.JSON.Encode.encode/3` directly. This will work for my purposes, but `Protobuf.JSON.Encode` feels like an intimate API, especially with `@moduledoc false`.

Some ideas on providing a more public API for getting the map representation of the JSON structure:

* Define `Protobuf.as_json/1` -- `as_json(struct) :: {:ok, map()}`

  Just like `Protobuf.encode/1`, and in this case it might also make sense to define `Protobuf.to_json/1`.

  BTW: I know there are plans to implement `to_json` in the included module, but it would also be good to have the option to call it in a generic place. I do this in a default phoenix view, like this:

  ```elixir
  def render("struct.proto", assigns) do
    assigns.struct
    |> Protobuf.encode()
  end

  def render("struct.json", assigns) do
    assigns.struct
    |> Azalea.ProtobufToJson.encode()
  end
  ```

* Another option, if you'd rather not expand the top-level public API, might be to add an option to `Protobuf.JSON.encode` that would allow the caller to decide it it should be encoded by the json library or not

  This feels similar to the `iolist` option that [can be provided to `Protobuf.Encoder.encode/2`](https://github.com/tony612/protobuf-elixir/blob/e1d4a23baa1f51c2da58db24f58a26dc4d6511b2/lib/protobuf/encoder.ex#L23).

## Thank You

Otherwise, this is looking great. Thanks for working on it!

britto:
Hey, @amiel. Thanks for your feedback.

If there are some good use cases for the map API, I think it will be OK to make it public. Do you need something similar (map->struct) for decoding too? I'm good with either `Protobuf.as_json` or something like `Protobuf.JSON.to_map`. @tony612, @whatyouhide any thoughts?

lib/protobuf/json.ex
britto: Just found out [it is possible](https://developers.google.com/protocol-buffers/docs/proto#using_proto3_message_types). I think the safest choice would be to move this check inside. Will do it.

amiel:
> Do you need something similar (map->struct) for decoding too?

Not actually at the moment, but I could definitely see using it. The reason I need `struct -> map` as opposed to `struct -> json` is that phoenix handles encoding json for me. I think it is common for web frameworks to do this automatically and I think it is common in web frameworks to think of native maps and arrays "as json". This also makes it easy to encode fragments "as json" so that they can be composed. It is hard to compose fragments of JSON that have already been encoded to JSON. I think this is why Rails has standardized on `#as_json`. Anyway, it would make sense to make that symetrical.

whatyouhide:
@britto on the first pass, I'm actually good with removing functions that work with JSON strings completely. We can have `Protobuf.JSON.to_encodable` and `Protobuf.JSON.from_decoded` which return and accept a "map" (or list, or integer, or any valid decoded JSON). This eliminates the dependency on Jason, too. Let's start with that and build on top of that, it should be pretty minimal changes :)

relistan:
Great start!

britto:
@amiel @whatyouhide done. I've exposed it as a `to_encodable/2` public API.

amiel:
@britto I'm doing some manual testing right now. I might be misunderstanding how this is supposed to work, but my assumption is that JSON keys would be underCamelCased and that's not what I'm seeing.

I have done the following, and I am still seeing under_scored keys in the generated JSON:

* I updated my mix dependencies to this git repo and this branch and update dependencies
* I re-ran `mix escript.install hex protobuf`
* I re-ran `protoc` for all of my `.proto` files

I did all this assuming that the new protoc plugin would update my .pb.ex definitions to include the `json_name` option. 

Thanks!

britto:
Hi @amiel. I think `mix escript.install hex protobuf` will actually grab the original binary from hex. Running `mix escript.install git https://github.com/britto/protobuf-elixir branch json-encoding` should do the trick.

amiel:
Yes, that worked. Thank you @britto !!!

amiel:
@britto Ok, integrating these changes with my application worked well.

The biggest difference is that my original code didn't omit defaults, so not all of my unit tests pass. I also didn't have `int64` serialized as a string before.

Anyway, this is working well.

britto:
@amiel awesome! do the tests pass with `emit_unpopulated` turned on?

amiel:
@britto YES! With one exception, which is my `int64` bug.

whatyouhide:
Awesome, thanks @britto!!!

amiel:
Will this get released soon? or will you wait until decoding is implemented too?

whatyouhide:
@amiel we'll wait until decoding is implemented. It might take a while for a release. If you need this, I recommend using the GitHub dependency!

amiel:
@whatyouhide Thanks. I've got my own internal implementation that is working for now. I'd like to switch to the official implementation as soon as it is released though :)

britto:
Just noticed something during my tests: `protoc` won't fail or warn when there are field name collisions ⚠️. I'll try to document that unexpected behavior somewhere later. This will affect decoding too.

https://github.com/protocolbuffers/protobuf/pull/5474

sleipnir:
Hello everyone, everything good? Excellent job with this feature, congratulations!
was this released? If not, what is the forecast for launch?

spokc:
Hey Guys, eagerly waiting for this

PR #109, opened 2020-06-18
New feature: JSON decoding

britto:
Addresses the second part of #86. Follow-up to #105.

This feature allows us to decode messages from JSON in addition to protobuf binaries.

It follows Google's specs and reference implementation. Only `proto3` syntax is supported at the moment and some features, such as well-known types, are not fully supported yet.

JSON decoding relies on the `Jason` library. Support to other JSON engines may be added in the future.

https://developers.google.com/protocol-buffers/docs/reference/google.protobuf

Huge thanks again to @whatyouhide and @amiel for their support and effort put into this.

### Design notes

This implementation iterates over the fields metadata to read from the JSON input. This means unknown fields will always be ignored and also means that it optimizes for "dense" JSON data. When the input is "sparse", containing very few fields set, the performance might be suboptimal.

### TODO

- [x] Find more edge cases to cover and test
- [x] Document functions and their options
- [x] Compare behavior against reference implementations (Go, Python, Java, C++)

### Out of scope

- Define `from_json` and `to_json` helpers for each generated module.
- `proto2` support
- `Any` mapping
- Google's well-known types mapping
- Handle duplicated map keys

### Known issues

According to the specs, libraries should raise an error on duplicated map keys. This is not possible at the moment, due to this limitation in Jason: https://github.com/michalmuskala/jason/issues/33

Field names might also collide and fail silently: https://github.com/protocolbuffers/protobuf/issues/5063. So extra care is advised when defining custom json field names.

### DIFF
--- a/lib/protobuf/dsl.ex
+++ b/lib/protobuf/dsl.ex
@@ -106,6 +106,12 @@ defmodule Protobuf.DSL do
     num_to_atom = for {fnum, %{name_atom: name_atom}} <- props, do: {fnum, name_atom}
     atom_to_num = for {name_atom, fnum, _opts} <- fields, do: {name_atom, fnum}, into: %{}
 
+    string_or_num_to_atom =
+      for {fnum, %{name: name, name_atom: name_atom}} <- props,
+          key <- [fnum, name],
+          do: {key, name_atom},
+          into: %{}
+
     Enum.map(atom_to_num, fn {name_atom, fnum} ->
       quote do
         def value(unquote(name_atom)), do: unquote(fnum)
@@ -124,6 +130,9 @@ defmodule Protobuf.DSL do
       [
         quote do
           def mapping(), do: unquote(Macro.escape(atom_to_num))
+        end,
+        quote do
+          def __reverse_mapping__(), do: unquote(Macro.escape(string_or_num_to_atom))
         end
       ]
   end
--- a/lib/protobuf/json.ex
+++ b/lib/protobuf/json.ex
@@ -53,6 +53,12 @@ defmodule Protobuf.JSON do
       iex> Protobuf.JSON.encode(message)
       {:ok, "{\\"color\\":\\"RED\\",\\"topSpeed\\":125.3}"}
 
+  And go the other way around with `decode/1`:
+
+      iex> json = ~S|{"color":"RED","topSpeed":125.3}|
+      iex> Protobuf.JSON.decode(json, Car)
+      {:ok, %Car{color: :RED, top_speed: 125.3}}
+
   JSON keys are encoded as camelCase strings by default, specified by the `json_name` field
   option. So make sure to *recompile the `.proto` files in your project* before working with
   JSON encoding, the compiler will generate all the required `json_name` options. You can set
@@ -63,9 +69,23 @@ defmodule Protobuf.JSON do
         double longitude = 2 [ json_name = "long" ];
       }
 
+  ## Known issues and limitations
+
+  Currently, the `protoc` compiler won't check for field name collisions, this library either.
+  So make sure your field names will be unique when serialized to JSON. For instance, this
+  message definition will not encode correctly, it will emit just one of the two fields, and
+  the problem might go unoticed:
+
+      message CollidingFields {
+        int32 f1 = 1 [json_name = "sameName"];
+        float f2 = 2 [json_name = "sameName"];
+      }
+
+  According to the specification, when duplicated JSON keys are found in maps, the library
+  should raise a decoding error. It currently ignores duplicates and keeps the last occurence.
   """
 
-  alias Protobuf.JSON.{Encode, EncodeError}
+  alias Protobuf.JSON.{Encode, EncodeError, Decode, DecodeError}
 
   @type encode_opt ::
           {:use_proto_names, boolean}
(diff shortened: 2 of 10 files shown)


amiel:
@britto I've got this running against my tests. I see this is still a draft. Please let me know when you are ready for feedback.

britto:
Hi @amiel! Thanks for the help. I've just pushed a new batch of unit tests and finished documenting. It's ready now.

lib/protobuf/json/decode.ex
```diff
@@ -0,0 +1,195 @@
...
+        value = field_value(prop, data) do
+      {prop.name_atom, decode_value(prop, value)}
+    end
+  end
+
+  defp decode_oneof_fields(data, %{field_props: field_props, oneof: oneofs}) do
+    for {oneof, index} <- oneofs,
+        {_field_num, %{oneof: ^index} = prop} <- field_props,
+        value = field_value(prop, data),
+        reduce: %{} do
```
britto, line 43: This is Elixir 1.8+ only. I can do it with a regular reduce if we need to support 1.7.

amiel:
@britto Ok, here are some thoughts:

* When there are multiple values for a oneof, this PR responds with an error

  My implementation goes with the last value. I don't have a strong opinion about this, but I thought I'd bring it up. I chose my implementation based on this piece of documentation:

  > If the parser encounters multiple members of the same oneof on the wire, only the last member seen is used in the parsed message.

  (found here https://developers.google.com/protocol-buffers/docs/proto3#oneof)

* Another case where this PR responds with an error and I'm not sure if it should is when an enum value is not expected

  My understanding is that it should be backwards compatible to remove enum values. For example, say an enum value is removed and the server starts using the updated protobuf structure, but a client is still on the old protobuf structure. In that case, the client is still sending that enum value, but the server should ignore it, as opposed to error. I could offer a more concrete example that would be helpful.

  What do you think?

* My first test had integers instead of symbols for enum values; I assume you've fixed that, since my tests are no longer showing that error

britto:
This is great. I've come to similar conclusions recently, but changed my mind after looking into the standard implementations. Let me try address these.

### Multiple fields set for a oneof

My initial implementation was ignoring  them too. But then – to my surprise – I found out the behavior in JSON is in fact different from the wire format behavior. Both Python and Go libs are expected to crash in that case:

https://github.com/protocolbuffers/protobuf/blob/f1ce8663ac88df54cf212d29ce5123b69203b135/python/google/protobuf/internal/json_format_test.py#L1047-L1054

https://github.com/protocolbuffers/protobuf-go/blob/bfc31022a5a55cbbfe0c712a541ca4231afc3abf/encoding/protojson/decode_test.go#L718-L724

### Invalid named enum value

As long as the enum values are encoded as integers, there is no problem, because proto3 accepts unkown integer values, they can even be negative integers. Both Python and Go respect that:

https://github.com/protocolbuffers/protobuf/blob/f1ce8663ac88df54cf212d29ce5123b69203b135/python/google/protobuf/internal/json_format_test.py#L890-L892

https://github.com/protocolbuffers/protobuf-go/blob/bfc31022a5a55cbbfe0c712a541ca4231afc3abf/encoding/protojson/decode_test.go#L454-L463

**However**... when we use the named version of those enums in JSON, the libraries will reject unknown names:

https://github.com/protocolbuffers/protobuf/blob/f1ce8663ac88df54cf212d29ce5123b69203b135/python/google/protobuf/internal/json_format_test.py#L886-L889

https://github.com/protocolbuffers/protobuf-go/blob/bfc31022a5a55cbbfe0c712a541ca4231afc3abf/encoding/protojson/decode_test.go#L472-L477

So I followed their lead. From Elixir's standpoint it even makes some sense, because accepting unknown named values would mean calling `String.to_atom/1` at some point, which is unsafe.

### Integers vs. Atoms

Exactly, I've changed it to always try to decode to an atom first, then keep it as an integer only if it's unknown.

lib/protobuf/json/decode.ex
```diff
@@ -0,0 +1,203 @@
...
+    with {:ok, integer} <- decode_integer(value),
+         true <- integer in @int_ranges[type] do
+      integer
+    else
+      _ -> throw({:bad_int, prop.name_atom, value})
+    end
+  end
+
+  defp decode_singular(%{type: type} = prop, value) when type in @float_types do
+    # TODO: validate int min-max range
```
britto, line 139: `double` fields will be automatically range-limited, because they are unrepresentable in Erlang already:

```elixir
iex> Jason.decode ~S|{"topSpeed":1.7976931348623159e+308}|
{:error,
 %Jason.DecodeError{
   data: "{\"topSpeed\":1.7976931348623159e+308}",
   position: 12,
   token: "1.7976931348623159e+308"
 }}
```

`float` fields however won't be automatically range-limited.

lib/protobuf/dsl.ex
```diff
@@ -123,14 +129,17 @@ defmodule Protobuf.DSL do
...
           def key(unquote(fnum)), do: unquote(name_atom)
         end
       end) ++
       [
         quote do
-          def mapping(), do: unquote(Macro.escape(mapping))
+          def mapping(), do: unquote(Macro.escape(atom_mapping))
+        end,
+        quote do
+          def reverse_mapping(), do: unquote(Macro.escape(reverse_mapping))
```
britto, line 142: Should probably name it `__reverse_mapping__/0` instead, since this is kind of private API.

lib/protobuf/json/decode.ex
```diff
@@ -0,0 +1,207 @@
...
+    message_props = Utils.message_props(module)
+    regular = decode_regular_fields(data, message_props)
+    oneofs = decode_oneof_fields(data, message_props)
+
+    module.__default_struct__()
+    |> struct(regular)
+    |> struct(oneofs)
+  end
+
+  def from_json_data(data, _module), do: throw({:bad_message, data})
```
whatyouhide, line 51: We'd have to do this here too if we go with the suggestion above:

```suggestion
  def from_json_data(data, module) when is_atom(module), do: throw({:bad_message, data})
```

lib/protobuf/json/decode.ex
```diff
@@ -0,0 +1,207 @@
...
+    for {_field_num, %{oneof: nil} = prop} <- field_props,
+        value = field_value(prop, data) do
+      {prop.name_atom, decode_value(prop, value)}
+    end
+  end
+
+  defp decode_oneof_fields(data, %{field_props: field_props, oneof: oneofs}) do
+    for {oneof, index} <- oneofs,
+        {_field_num, %{oneof: ^index} = prop} <- field_props,
+        not is_nil(value = field_value(prop, data)) do
```
whatyouhide, line 63: Wdyt about

```suggestion
    for {oneof, index} <- oneofs,
        {_field_num, %{oneof: ^index} = prop} <- field_props,
        value = field_value(prop, data),
        not is_nil(value) do
```

lib/protobuf/json/decode.ex
```diff
@@ -0,0 +1,207 @@
...
+    fixed64: @int64_range,
+    uint32: @uint32_range,
+    uint64: @uint64_range
+  }
+
+  @int_types Map.keys(@int_ranges)
+
+  @float_types [:float, :double]
+
+  def from_json_data(data, module) when is_map(data) do
```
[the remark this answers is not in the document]
britto: Good catch.

lib/protobuf/json/decode.ex
britto: Unfortunately, this breaks test/protobuf/json/decode_test.exs:487 💔 because when `value` is `false` it won't run the separate `is_nil` check.

whatyouhide: Oh shit yuck 😄 Okay, let's figure this out later, no worries!

lib/protobuf/json.ex
```diff
@@ -166,4 +186,93 @@ defmodule Protobuf.JSON do
...
+      iex> ~S|{"color":"RED"}| |> Protobuf.JSON.decode(Car)
+      {:ok, %Car{color: :RED, top_speed: 0.0}}
+
+      iex> ~S|{"color":"GREEN","topSpeed":80.0}| |> Protobuf.JSON.decode(Car)
+      {:ok, %Car{color: :GREEN, top_speed: 80.0}}
+  """
+  @spec decode(iodata, module) :: {:ok, struct} | {:error, DecodeError.t() | Exception.t()}
+  def decode(iodata, module) do
+    if Code.ensure_loaded?(Jason) do
+      with {:ok, json_data} <- Jason.decode(iodata),
```
britto, line 248: Just noticed this triggers a compilation warning on the `bench` project, which doesn't have `jason` in its deps.

```
==> protobuf
Compiling 33 files (.ex)
warning: Jason.decode/1 is undefined (module Jason is not available or is yet to be defined)
  lib/protobuf/json.ex:248: Protobuf.JSON.decode/2

warning: Jason.encode/1 is undefined (module Jason is not available or is yet to be defined)
  lib/protobuf/json.ex:161: Protobuf.JSON.encode/2
```

We'd probably need to conceal it in a variable to avoid it. I'll take a look.

whatyouhide: Yeah this is going to work:

```elixir
jason = Jason

if Code.ensure_loaded?(jason) do
  jason.decode(iodata)
```

:)

lib/protobuf/json/decode.ex
britto: In `decode_regular_fields/2` this is not a problem because we can safely skip all _falsey_ values. For oneof fields we need this. We can probably make some changes to `field_value/2` to make this prettier though. Will see.

lib/protobuf/json.ex
britto: Fixed with a private helper 👍

whatyouhide:
Awesome, thanks so much @britto 💟

PR #110, opened 2020-06-18
Update Elixir, Erlang and Protobuf

britto:
* Update the CI matrix
* Bump Elixir and Erlang
* ~Drop Elixir 1.7 and Erlang 20.0 support~
* Bump Protobuf and recompile internal protos

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ install:
   - mix deps.get
   - mix eqc.install --mini --force
 before_script:
-  - wget https://github.com/protocolbuffers/protobuf/releases/download/v3.11.2/protoc-3.11.2-linux-x86_64.zip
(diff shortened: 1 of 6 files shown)


mix.exs
```diff
@@ -7,7 +7,7 @@ defmodule Protobuf.Mixfile do
     [
       app: :protobuf,
       version: @version,
-      elixir: "~> 1.6",
```
whatyouhide, line 10: If possible, let's not bump the version yet. I don't see a strong enough reason yet :)

britto: Good point. I've reverted the changes in #109 that required 1.8+ too.

britto: Done.

whatyouhide:
Awesome!

PR #115, opened 2020-06-26
Fix extensions generator output

britto:
Make sure it renders correctly when the list of extensions is long.


whatyouhide:
Good catch!

PR #117, opened 2020-06-30
Improve benchmarks

britto:
This changes the way we benchmark the library code, making it easier to compare different variations of the same decode or decode functions across different branches. In addition to that, it also adds:

* Elixir modules for the remaining canonical benchmarks (3 and 4)
* A make task to regenerate message modules
* A script to download datasets 3 and 4 (basically the same as the [original](https://github.com/protocolbuffers/protobuf/blob/master/benchmarks/download_data.sh))


whatyouhide:
Looks great 👍

PR #118, opened 2020-07-01
Leverage iodata on encoding

britto:
Reducing the amount of intermediate binaries allocated speeds up encoding and potentially decreases memory usage.

I'm usually very distrustful of local benchmarks, so it'd be very reassuring to see some production numbers with these optimizations applied. Anyway, these are the results I've got running against `master`:

```
##### With input google_message1_proto3 #####
Name                                                ips        average  deviation         median         99th %
encode (optimized)                              24.24 K       41.26 μs  ±4792.01%           0 μs           0 μs
encode (master)                                 10.70 K       93.47 μs  ±2984.15%          12 μs          32 μs

Comparison: 
encode (optimized)                              24.24 K
encode (master)                                 10.70 K - 2.27x slower +52.21 μs

Memory usage statistics:

Name                                         Memory usage
encode (optimized)                                2.78 KB
encode (master)                                   3.55 KB - 1.28x memory usage +0.77 KB

**All measurements for memory usage were the same**

##### With input google_message1_proto2 #####
Name                                                ips        average  deviation         median         99th %
encode (optimized)                              13.94 K       71.71 μs  ±3642.67%           0 μs           0 μs
encode (master)                                  6.28 K      159.30 μs  ±2283.59%          22 μs          77 μs

Comparison: 
encode (optimized)                              13.94 K
encode (master)                                  6.28 K - 3.86x slower +118.05 μs

Memory usage statistics:

Name                                         Memory usage
encode (optimized)                                7.49 KB
encode (master)                                   9.59 KB - 3.45x memory usage +6.80 KB

**All measurements for memory usage were the same**

##### With input google_message3_3 #####
Name                                                ips        average  deviation         median         99th %
encode (optimized)                              13.63 K       73.38 μs  ±3569.64%           0 μs           0 μs
encode (master)                                  6.05 K      165.20 μs  ±2330.28%           6 μs       18.76 μs

Comparison: 
encode (optimized)                              13.63 K
encode (master)                                  6.05 K - 4.00x slower +123.95 μs

Memory usage statistics:

Name                                         Memory usage
encode (optimized)                                1.82 KB
encode (master)                                   2.38 KB - 0.85x memory usage -0.40625 KB

**All measurements for memory usage were the same**

##### With input google_message3_5 #####
Name                                                ips        average  deviation         median         99th %
encode (optimized)                               8.36 K      119.66 μs  ±2753.18%           0 μs           0 μs
[...]

PR #119, opened 2020-07-01
Tweak benchmark scripts

britto:
Separating part of #118 into its own PR to make local testing easier for everyone.

* Increase run time for more accuracy
* Enable memory measurements
* Bring back `google_message3_1`, even though it may go over the established run time
* Select the longest payload rather than the first one
* Output comparison to console in addition to HTML


whatyouhide:
Awesome!

PR #126, opened 2020-08-03
Ugly fix for multiple options in submessage

shengwu:


shengwu:
Whoops, didn't mean to make a pull request to the parent repo. 

I ran into an issue with using multiple field options on a field:
```
message Example {
  string foo = 1 [
    (ext.a) = true,
    (ext.b) = "bob"
  ];
}
```

I can file an issue about this instead of trying to contribute - I'm not super experienced w/elixir.

PR #128, opened 2020-08-06
Explicitly set default value of enum should encode in proto2

xinz:


test/protobuf/encoder_test.exs
```diff
@@ -169,16 +169,44 @@ defmodule Protobuf.EncoderTest do
...
 
-  test "encodes default value for proto2" do
+  test "encodes enum default value for proto2" do
     # Includes required
-    msg = TestMsg.Bar2.new(a: 0)
+    msg = TestMsg.EnumBar2.new(a: 0)
     assert Encoder.encode(msg) == <<8, 0>>
 
-    # Excludes optionals at default value
-    msg = TestMsg.Bar2.new(a: 0, b: 0)
```
xinz, line 178: The original assertion
```
msg = TestMsg.Bar2.new(a: 0, b: 0)
assert Encoder.encode(msg) == <<8, 0>>
```
The `msg` after encoding should be `<<8, 0, 16, 0>>`, please see the related [issue](https://github.com/tony612/protobuf-elixir/issues/129) for comparing.

PR #136, opened 2020-11-15
[fix warn] use __STACKTRACE__ and update deps

redink:


mix.exs
```diff
@@ -22,7 +22,7 @@ defmodule Protobuf.Mixfile do
   def application do
     [
       mod: {Protobuf.Application, []},
-      extra_applications: [:logger]
+      extra_applications: [:logger, :eex]
```
redink, line 25:
```
warning: EEx.compile_file/2 defined in application :eex is used by the current application but the current application does not directly depend on :eex. To fix this, you must do one of:

  1. If :eex is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs

  2. If :eex is a dependency, make sure it is listed under "def deps" in your mix.exs

  3. In case you don't want to add a requirement to :eex, you may optionally skip this warning by adding [xref: [exclude: EEx] to your "def project" in mix.exs

Found at 3 locations:
  lib/protobuf/protoc/template.ex:8: Protobuf.Protoc.Template
  lib/protobuf/protoc/template.ex:16: Protobuf.Protoc.Template
  lib/protobuf/protoc/template.ex:20: Protobuf.Protoc.Template

warning: EEx.function_from_file/5 defined in application :eex is used by the current application but the current application does not directly depend on :eex. To fix this, you must do one of:

  1. If :eex is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs

  2. If :eex is a dependency, make sure it is listed under "def deps" in your mix.exs

  3. In case you don't want to add a requirement to :eex, you may optionally skip this warning by adding [xref: [exclude: EEx] to your "def project" in mix.exs

Found at 3 locations:
  lib/protobuf/protoc/template.ex:8: Protobuf.Protoc.Template
  lib/protobuf/protoc/template.ex:16: Protobuf.Protoc.Template
  lib/protobuf/protoc/template.ex:20: Protobuf.Protoc.Template

Generated protobuf app
```

whatyouhide (approved):

tony612 (approved):

PR #140, opened 2020-12-23
Fix repeated enum field generate [[]] into typespec

### LINKED ISSUE #132: Repeated enum fields do not generate proper typespecs
I'm not 100% sure about this sorry so this is more a question than an statement

In the case we have

```protobuf
syntax = "proto3";
package test;

message Salute {
    repeated Word words = 1;
}

enum Word {
  WORD_HI = 0;
  WORD_BYE = 1;
  WORD_MISS = 2;
  WORD_MR = 3;
}
```

The generate generates the following

```elixir
defmodule Test.Salute do
  @moduledoc false
  use Protobuf, syntax: :proto3

  @type t :: %__MODULE__{
    words: [[Test.Word.t()]],
  }

  defstruct [
    :words
  ]

  field :words, 0, repeated: true, type: Test.Word, enum: true
end

defmodule Test.Word do
  @moduledoc false
  use Protobuf, enum: true, syntax: :proto3

  @type t ::
          integer
          | :WORD_HI
          | :WORD_BYE
          | :WORD_MISS
          | :WORD_MR

  field :WORD_HI, 0
  field :WORD_BYE, 1
  field :WORD_MISS, 2
  field :WORD_MR, 3
end
```

Would you mind to confirm that is the expected typespec instead just `words: [Test.Word.t()],` so just one single `[` instead of `[[`

wingyplus:
Removing `[]` out of `Protobuf.Protoc.Generator.Message.fmt_type/1` and
add `[]` into `Protobuf.TypeUtil.enum_to_spec/3` when `TYPE_MESSAGE` is match
alongs with repeated true.

Fixes #132

### DIFF
--- a/lib/protobuf/protoc/generator/message.ex
+++ b/lib/protobuf/protoc/generator/message.ex
@@ -154,7 +154,7 @@ defmodule Protobuf.Protoc.Generator.Message do
   end
 
   defp fmt_type(%{label: "repeated", type_enum: type_enum, type: type}) do
-    "#{type_to_spec(type_enum, type, true)}"
+    type_to_spec(type_enum, type, true)
   end
 
   defp fmt_type(%{type_enum: type_enum, type: type}) do
--- a/lib/protobuf/protoc/generator/message.ex
+++ b/lib/protobuf/protoc/generator/message.ex
@@ -115,6 +115,7 @@ defmodule Protobuf.Protoc.Generator.Message do
         {fmt_type_name(f.name, longest_width), "{atom, any}"}
       end)
 
+
     types =
       types ++
         Enum.map(fields, fn f ->
@@ -153,7 +154,7 @@ defmodule Protobuf.Protoc.Generator.Message do
   end
 
   defp fmt_type(%{label: "repeated", type_enum: type_enum, type: type}) do
-    "[#{type_to_spec(type_enum, type, true)}]"
+    "#{type_to_spec(type_enum, type, true)}"
   end
 
   defp fmt_type(%{type_enum: type_enum, type: type}) do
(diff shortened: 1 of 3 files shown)


lib/protobuf/type_util.ex
```diff
@@ -36,7 +36,7 @@ defmodule Protobuf.TypeUtil do
   def enum_to_spec(:TYPE_SINT32), do: "integer"
   def enum_to_spec(:TYPE_SINT64), do: "integer"
   def enum_to_spec(_), do: "any"
-  def enum_to_spec(:TYPE_MESSAGE, type, true = _repeated), do: "#{type}.t"
+  def enum_to_spec(:TYPE_MESSAGE, type, true = _repeated), do: "[#{type}.t]"
```
wingyplus, line 39: I just see that message type doesn't add `| nil` into generated type. Should we need to add into this patch?

lib/protobuf/protoc/generator/message.ex
```diff
@@ -153,7 +154,7 @@ defmodule Protobuf.Protoc.Generator.Message do
   end
 
   defp fmt_type(%{label: "repeated", type_enum: type_enum, type: type}) do
-    "[#{type_to_spec(type_enum, type, true)}]"
+    "#{type_to_spec(type_enum, type, true)}"
```
whatyouhide, line 157: Can this be 👇 now?

```suggestion
    type_to_spec(type_enum, type, true)
```

whatyouhide (commented):
Hey @wingyplus , sorry for the long wait! Could you rebase this on the master branch?

lib/protobuf/protoc/generator/message.ex
wingyplus: @whatyouhide This comment is apply. :)

wingyplus:
@whatyouhide thank you for reviewing my changes. The PR already rebase and apply your suggestion. Please kindly review again.

PR #149, opened 2021-04-16
Replace eqc with stream_data

whatyouhide:
Reasons:

1. stream_data is faster: one of the test files went from 0.6s to 0.2s, the other from 0.5s to 0.2s when switching to stream_data
2. We don't need the additional `mix eqc.install --mini` step in order to run the test suite

Thoughts @tony612?


whatyouhide:
Actually, let me for once take the liberty to apologize rather than ask for permissions 😄, I'll merge this and move on so that I can also setup GitHub Actions to run CI (#148). @tony612 please let me know if this causes any problems and we can always revert or revert in part!

PR #151, opened 2021-04-17
Upgrade Elixir from 1.11.2 to 1.11.4 in GitHub workflow

wingyplus:


whatyouhide:
Nice catch, thanks @wingyplus! 💟

PR #158, opened 2021-06-08
Fix compiler warning for Elixir versions prior to 1.12

britto:
That warning no longer shows up in Elixir 1.12, but still does in 1.11 and lower. Adding `generated: true` in a few places fixes it.

    warning: the result of the expression is ignored (suppress the warning by assigning the expression to the _ variable)
      lib/protobuf/decoder.ex:103

### DIFF
--- a/lib/protobuf/wire/varint.ex
+++ b/lib/protobuf/wire/varint.ex
@@ -48,6 +48,8 @@ defmodule Protobuf.Wire.Varint do
   @max_bits 64
   @mask64 bsl(1, @max_bits) - 1
 
+  # generated: true is required here to silence compilation warnings in Elixir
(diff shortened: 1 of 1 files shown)


britto:
Oddly though, the warnings are gone in 1.10 and 1.11 but still show up in 1.8 and 1.9 🤔

britto:
@ericmj that's a good idea! I will install 1.8 and 1.9 locally tomorrow and try to further narrow it down 🔍

britto:
@ericmj here's a breakdown of my findings per version:

* `1.12`: no warnings, no changes required.
* `1.11` and `1.10`: behave the same, this is the minimal set of `generated: true` required to silence the warnings.
* `1.9` and `1.8`: also behave the same, unable to fix even when adding `generated: true` to all `quote` calls.

PR #159, opened 2021-06-08
CI improvements

britto:
* Add Elixir 1.12 and OTP 24 to the build matrix.
* Drop Elixir 1.8 and OTP 21.
* Remove deprecated Travis CI config.

Also make sure the tests actually run in CI. The previous setup was skipping all the tests due to a combination of `--only integration` in the command line and `exclude: [integration: true]` in `test_helper.exs`. Now we are running all regular tests in a step:

    23 doctests, 13 properties, 298 tests, 0 failures, 4 excluded, 3 skipped

Then the 4 integration ones in another step, after we install `protoc` and make sure it compiles a few files correctly, as before:

    23 doctests, 13 properties, 298 tests, 0 failures, 330 excluded


.github/workflows/main.yml
```diff
@@ -43,4 +45,21 @@ jobs:
...
+
+      - name: Install protoc
+        if: ${{matrix.integration}}
+        run: |
+          wget https://github.com/protocolbuffers/protobuf/releases/download/v3.12.3/protoc-3.12.3-linux-x86_64.zip
+          unzip protoc-3.12.3-linux-x86_64.zip
+          echo "./bin" >> $GITHUB_PATH
+
+      - name: Compile .proto files to Elixir with protoc
+        if: ${{matrix.integration}}
```
britto, line 58: I don't know if we should actually flag it like this or move it to a separate `job`.

.travis.yml
```diff
@@ -1,35 +0,0 @@
...
-before_script:
-  - wget https://github.com/protocolbuffers/protobuf/releases/download/v3.12.3/protoc-3.12.3-linux-x86_64.zip
-  - unzip protoc-3.12.3-linux-x86_64.zip
-  - export PATH=$PATH:$PWD/bin/
-script:
-  - mix test
-  - mix test.integration
-  - mix hex.build
-  # Only run for latest version
-  - '[[ "$TRAVIS_ELIXIR_VERSION" =~ 1.[8-9] ]] || mix format --check-formatted'
```
britto, line 35: I noticed we don't check formatting in the new pipeline anymore. Should we bring it back?

.github/workflows/main.yml
```diff
@@ -43,4 +45,21 @@ jobs:
...
 
       - name: Run tests
-        run: mix test.integration --trace
+        run: mix test --trace
+
+      - name: Install protoc
+        if: ${{matrix.integration}}
+        run: |
+          wget https://github.com/protocolbuffers/protobuf/releases/download/v3.12.3/protoc-3.12.3-linux-x86_64.zip
+          unzip protoc-3.12.3-linux-x86_64.zip
```
britto, line 54: Even though it doesn't take too long, it would be ideal if we could cache this. Maybe an improvement for the next iteration.

.github/workflows/main.yml
whatyouhide: I think this is great 👍

whatyouhide (approved):
Lovely 💯

.travis.yml
whatyouhide: Let's do it yes, but in a later PR :)

whatyouhide: @britto actually can we also add `mix compile --warnings-as-errors` and `mix deps.unlock --check-unused` to the `lint: true` CI step as part of linting?

britto: Cool, I'll give it a shot.

PR #164, opened 2021-07-13
Ensure proper cleanup on application shutdown

### LINKED ISSUE #130: Application start is not idempotent
As @igormq reported in gRPC repository [here](https://github.com/elixir-grpc/grpc/issues/165) (but it is actually a problem with this library), starting `:protobuf` is not idempotent, which means that if you need to restart the application (which can happen on purpose on some environments, such as how Nerve's Shoehorn works) protobuf refuses to start again.

This is due to this code here: https://github.com/tony612/protobuf-elixir/blob/master/lib/protobuf/extension.ex#L134-L136

I'm not sure why that is there, but I think we need to rethink/remove that.

In order to reproduce the problem, just do:

```elixir
Application.ensure_started(:protobuf)
Application.stop(:protobuf)
Application.ensure_started(:protobuf)
```

The expected behavior would be to protobuf app to just start, but since it is using persistent_term to store and it is crashing if it is set, then it cause the application to throw and not start, requiring a full VM restart (or unsetting the persistent_term by hand, which seems like a hack messing with protobuf internals).

I can contribute with a fix, but I first need to understand why that is there and discuss approaches to solve that.

Thanks

britto:
This will leave the BEAM in a clean state once the app shuts down by dropping all extensions currently loaded. As a consequence, this also ensures clean restarts. It closes #145, as an alternative approach to fix #130.

Since we dropped support to OTP 21, I also seized the opportunity to remove the `GlobalStore` wrapper around `:persistent_term`.


lib/protobuf/extension.ex
```diff
@@ -131,12 +129,18 @@ defmodule Protobuf.Extension do
...
           raise "Extension #{inspect(ext.extendee)}##{fnum} already exists"
         end
 
-        GlobalStore.put(fnum_key, mod)
+        :persistent_term.put(fnum_key, mod)
       end)
     end)
   end
+
+  def unload_extensions do
```
whatyouhide, line 141: **chore (blocking)**: If this is public, then we need to add a spec and docs. If it's only meant for testing, then let's add `@doc false` and a commend saying `# Made public for testing`.

whatyouhide (changes requested):
Looking great! One question I have is: if we rely on `:persistent_term` directly, isn't this a breaking change?

lib/protobuf/extension.ex
britto: Good catch! I believe visibility is public here not because of testing but actually because we call these functions (both `cal_extensions/1` and `unload_extensions/0`) from `Protobuf.Application`. Maybe we should rename them to `__cal_extensions__/1` and `__unload_extensions/0` too, to make it clearer it is part of a private API?

whatyouhide: Yes, let's rename to `__...__`, make them `@doc false` 👍

britto:
> Looking great! One question I have is: if we rely on :persistent_term directly, isn't this a breaking change?

It's a breaking change for apps running on OTP 21.1 and lower.

whatyouhide (approved):

jeanparpaillon:
:+1:

PR #172, opened 2021-10-14
Add repeated scalar types to property-based roundtrip tests

britto:
Follow-up to #169. These tests fail on master – further confirming the bug fixed in there – and will pass once we rebase against it.


britto:
Rebased ✅

PR #217, opened 2021-11-29
Fix enum aliasing

### LINKED ISSUE #214: Fix enum aliases
See <https://developers.google.com/protocol-buffers/docs/proto>:

Screenshot 2021-11-29 at 09 23 50

whatyouhide:
Closes #214.


lib/protobuf/dsl.ex
```diff
@@ -100,11 +101,11 @@ defmodule Protobuf.DSL do
     num_to_atom = for {fnum, %{name_atom: name_atom}} <- props, do: {fnum, name_atom}
     atom_to_num = for {name_atom, fnum, _opts} <- fields, do: {name_atom, fnum}, into: %{}
 
-    string_or_num_to_atom =
-      for {fnum, %{name: name, name_atom: name_atom}} <- props,
-          key <- [fnum, name],
-          do: {key, name_atom},
-          into: %{}
+    reverse_mapping =
+      for {name_atom, field_number, _opts} <- fields,
```
whatyouhide, line 105: Using the DSL fields instead of the `MessageProps` here. We probably want to fix the message props in the long run, but this works for now (and we use the fields already, so I'm not doing anything we weren't already doing).

PR #219, opened 2021-11-29
Skip encoding for matching proto2 defaults

whatyouhide:
conformance tests go brrrrr


lib/protobuf/encoder.ex
```diff
@@ -78,6 +78,7 @@ defmodule Protobuf.Encoder do
   defp skip_field?(_syntax, [], _prop), do: true
   defp skip_field?(_syntax, val, _prop) when is_map(val), do: map_size(val) == 0
   defp skip_field?(:proto2, nil, %FieldProps{optional?: optional?}), do: optional?
+  defp skip_field?(:proto2, value, %FieldProps{default: value}), do: true
```
whatyouhide, line 81: 😑

britto (commented):
Great catch! Thanks! 💜

PR #240, opened 2021-12-21
Don't skip storing the value of a oneof field even if it's the default value

### LINKED ISSUE #239: oneof fields incorrectly skipped when value matches the default
There is a [rule](https://github.com/elixir-protobuf/protobuf/blob/6b8f40f206375c36666f2cc9a570a061f9f0d89a/lib/protobuf/encoder.ex#L78) in `encoder.ex`to skip encoding of a field value if it matches the field's default value. This is incorrect in the case of `oneof` fields, since the explicit value lets us determine which case was set. Note the the [language guide](https://developers.google.com/protocol-buffers/docs/proto#oneof) also indicates that the field should be set in this case:
> If you set a oneof field to the default value (such as setting an int32 oneof field to 0), the "case" of that oneof field will be set, and the value will be serialized on the wire.

The current behavior is that the field is skipped, and then becomes blank on decoding. I have a simple fix that I've provided as a PR.

mikebveil:
Resolves #239.

### DIFF
--- a/lib/protobuf/encoder.ex
+++ b/lib/protobuf/encoder.ex
@@ -75,7 +75,7 @@ defmodule Protobuf.Encoder do
   defp skip_field?(_syntax, [], _prop), do: true
   defp skip_field?(_syntax, val, _prop) when is_map(val), do: map_size(val) == 0
   defp skip_field?(:proto2, nil, %FieldProps{optional?: optional?}), do: optional?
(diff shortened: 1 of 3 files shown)


whatyouhide:
@mikebveil this is fantastic, good catch! Before merging, can you please add a test case for this?

mikebveil:
Should be all set now.

ericmj (approved):

PR #246, opened 2022-01-04
Fix compilation warning for Elixir 1.9.4

zolakeith:
Compilation fails with 1.9.4 using the `--warnings-as-errors` with the
warning that "the result of the expression is ignored". This commit
simply assigns the result to `_` to suppress the warning.

Note: This warning doesn't appear in your CI but it does show on [this page](https://github.com/elixir-protobuf/protobuf/commit/fe11c65da7f8316faa01c4f4336557a8708b7a6e#diff-572680e4c3bdfcb24caac848aefec9c3065cdaa9d43c60d9dd343eb57d2cc4dd). It blocked our fork, hence the fix, but maybe you've already worked around it somehow?

### DIFF
--- a/lib/protobuf/wire/varint.ex
+++ b/lib/protobuf/wire/varint.ex
@@ -138,7 +138,7 @@ defmodule Protobuf.Wire.Varint do
           x6::7, 1::1, x7::7, 1::1, x8::7, 0::1, x9::7>>
       end,
       quote(generated: true) do
-        band(
+        v =
           x0 +
             bsl(x1, 7) +
             bsl(x2, 14) +
@@ -148,9 +148,9 @@ defmodule Protobuf.Wire.Varint do
             bsl(x6, 42) +
             bsl(x7, 49) +
             bsl(x8, 56) +
-            bsl(x9, 63),
-          unquote(@mask64)
-        )
+            bsl(x9, 63)
+
+        _ = band(v, unquote(@mask64))
       end
     }
   ]


ericmj:
We usually only run `--warnings-as-errors` on the latest Elixir version in CI for this reason. Can you do the same in your CI since this is an Elixir issue and not an issue with the library?

lib/protobuf/wire/varint.ex
```diff
@@ -138,7 +138,7 @@ defmodule Protobuf.Wire.Varint do
           x6::7, 1::1, x7::7, 1::1, x8::7, 0::1, x9::7>>
       end,
       quote(generated: true) do
-        band(
+        v =
```
whatyouhide, line 141: Would doing `_ = ` here directly still fix the warning?

lib/protobuf/wire/varint.ex
```diff
@@ -163,7 +163,7 @@ defmodule Protobuf.Wire.Varint do
 
   defp def_decoder_success_clauses(name, args, body) do
     for {pattern, expression} <- @varints do
-      quote do
+      quote location: :keep do
```
whatyouhide, line 166: Why are we changing this?

lib/protobuf/wire/varint.ex
zolakeith: I did try that, but it didn't work!

lib/protobuf/wire/varint.ex
zolakeith: I changed that to find where the problem was - the message was just saying this:
warning
 
 Which just points to [this line](https://github.com/elixir-protobuf/protobuf/blob/main/lib/protobuf/decoder.ex#L121). I'm happy to revert that if you prefer?

zolakeith:
@ericmj I'm seeing the warning on this repo too. I attached a pic in my response to @whatyouhide :point_up:

lib/protobuf/wire/varint.ex
whatyouhide: Cool if this change fixes the warning, Im good with it! Does it? 😄

britto (approved):
Awesome! Thanks for fixing that, @zolakeith.

whatyouhide:
Thanks @zolakeith! 💟

PR #249, opened 2022-01-05
Add mix .proto compiler

jeanparpaillon:
This PR adds a compiler for `.proto` files as a mix compiler.

It prevents version mismatch between `protoc-gen-elixir` and protobuf lib


whatyouhide:
Hey @jeanparpaillon, thanks for the PR! I'm not really following what the purpose of this is.

You mention version mismatch. Are you talking about the fact that you could have some generated files that were generated with version `x` of `protoc-gen-elixir`, but are then compiled and loaded by your application when the application is depending on version `y` of this library?

jeanparpaillon:
Hi @whatyouhide I've updated the PR with improved doc 
You can find a more complete rationale in the commit message and this PR's description

README.md
```diff
@@ -63,12 +63,28 @@ end
...
+  ```elixir
+  def project do
+    [
+      ...
+      compilers: [:proto | Mix.compilers()]
+      ...
+    ]
+  end
+  ```
+
```
zolakeith, line 85:
```suggestion

  Your `.proto` files will be compiled whenever you type `mix compile` or `iex -S mix`.
```

zolakeith: Actually this would be a good place to explain to the user the files will only be re-compiled if there is any change to the .proto file, or the config in mix.exs is changed

jeanparpaillon: updated README.md

whatyouhide:
@jeanparpaillon my biggest issue with this is that it's not solving the problem of version mismatch.

It is only solving that problem for the subset of users that keep `.proto` files definition inside the Elixir project they're also compiling them in. I have no data on what approach is the most common, but anecdotally where I work we keep the `.proto` source separate. I wrote [a blog post](https://andrealeopardi.com/posts/sharing-protobuf-schemas-across-services/) about that. I feel like that approach rides on the benefit that Protobuf is a language-agnostic protocol.

Anyways, for users that keep the `.proto` file generation and usage separate, version mismatch will still be a problem. The solution to that problem, in my opinion, is to store the version of this library used to generate `.pb.ex` files in those files themselves. Something along the lines of:

```elixir
defmodule MyMessage do
  use Protobuf, ..., protoc_gen_elixir_version: "0.9.1"
end
```

We can then use semantic versioning properties to check whether versions are compatible. We can do that at compile time when compiling the `.pb.ex` files.

### `.proto` compiler

For folks that keep Protobuf schemas in their Elixir project, I think this compiler is a pretty nice feature though. However, it's not really using any of the code in this library, which makes me think it could be build as a separate library (`protobuf_compiler`) that folks with this use case bring into their project. What are your thoughts on that? We can point to that library from the README and documentation of this repo. We might even host it in the elixir-protobuf GitHub org.

jeanparpaillon:
@whatyouhide Hi! Based on your suggestion, I've moved the code to a separate repo: https://github.com/OffgridElectric/protobuf_compiler

Thank you for pointing out good practices (proto files in a shared/separate code repo).

I am ok to transfer ownership to elixir-protobuf, as well as improving plugin version checking. I would separate the 2 concerns:
* on protobuf side: add a mechanism to ensure generated code is compatible with the library
* on compiler side, help compiling with the right version of the plugin, based on user requirements.

PR #253, opened 2022-01-07
Fix FunctionClauseError in in Protobuf.Protoc.CLI.parse_params/2

amandasposito:
I started to get this error after I upgraded the protobuf lib and tried to generate the proto files passing no arguments.

I'm using Elixir 1.11.2 and Erlang/OTP 23.

```elixir
** (FunctionClauseError) no function clause matching in Protobuf.Protoc.CLI.parse_params/2

    The following arguments were given to Protobuf.Protoc.CLI.parse_params/2:

        # 1
        %Protobuf.Protoc.Context{custom_file_options: %{}, dep_type_mapping: %{}, gen_descriptors?: false, global_type_mapping: %{}, module_prefix: nil, namespace: [], one_file_per_module?: false, package: nil, package_prefix: nil, plugins: [], syntax: nil, transform_module: nil}

        # 2
        nil

    (protobuf 0.9.0) lib/protobuf/protoc/cli.ex:73: Protobuf.Protoc.CLI.parse_params/2
    (protobuf 0.9.0) lib/protobuf/protoc/cli.ex:51: Protobuf.Protoc.CLI.main/1
    (elixir 1.11.2) lib/kernel/cli.ex:124: anonymous fn/3 in Kernel.CLI.exec_fun/2
--elixir_out: protoc-gen-elixir: Plugin failed with status code 1.
```

### DIFF
--- a/lib/protobuf/protoc/cli.ex
+++ b/lib/protobuf/protoc/cli.ex
@@ -48,7 +48,7 @@ defmodule Protobuf.Protoc.CLI do
 
     ctx =
       %Context{}
-      |> parse_params(request.parameter || "")
+      |> parse_params(request.parameter)
       |> find_types(request.proto_file)
 
     files =
@@ -68,6 +68,8 @@ defmodule Protobuf.Protoc.CLI do
 
   # Made public for testing.
   @doc false
+  def parse_params(%Context{} = ctx, nil), do: ctx
+
   def parse_params(%Context{} = ctx, params_str) when is_binary(params_str) do
     params_str
     |> String.split(",")
--- a/test/protobuf/protoc/cli_test.exs
+++ b/test/protobuf/protoc/cli_test.exs
@@ -79,6 +79,12 @@ defmodule Protobuf.Protoc.CLITest do
         parse_params(%Context{}, "package_prefix=,gen_descriptors=true")
       end
     end
+
+    test "won't raise when there are no arguments" do
+      assert parse_params(%Context{}, "") == %Context{}
+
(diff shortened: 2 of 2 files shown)


whatyouhide:
Hey @amandasposito, thanks for the contribution! 💟 

I don't understand where the `parse_params(ctx, nil)` call was coming from given that we were doing `|| ""` when calling it from the `CLI.main` function. 🤔

amandasposito:
Hey @whatyouhide, thanks for taking a look at it.

I was talking to @noteu about this bug, trying to identify the root cause of this, and I think we figure it out.

One thing that I didn't mention is that I use `protobuf` via `twirp`. What seems to be happening is that `twirp` uses the (not-so) internal API, calling the `parse_params` directly. 
https://github.com/keathley/twirp-elixir/blob/master/lib/twirp/protoc/cli.ex#L16

After the CLI changes, this error started to raise.
https://github.com/elixir-protobuf/protobuf/commit/5f9f0e07db018d0e7df7b6bbba2a705d238f8b89

I'm okay fixing it on the twirp side and closing this PR. The only thing is that since this is a public function it could happen that someone does that again.

whatyouhide:
Hey @amandasposito, João reached out too, fantastic!

So, `parse_params/2` has `@doc false` so while it's a public function in the code itself, it's not part of the public API. This means that yes, this is something that needs to be fixed in Twirp so as not to use private APIs like `parse_params/2`.

Hope that's a good solution for you folks 🙃 

Thanks anyways for reporting this and taking the time to fix it 💟

PR #258, opened 2022-01-14
Suffix version with `-dev` when not released

### LINKED ISSUE #256: Plugin version
Plugin version should be more precise

When building plugin from sources, it reports a fixed version (from `mix.exs`). Sthg like `git describe` should be used to avoid confusion.
As of today, library from HEAD is not compatible with code generated with plugin 0.9.0 (released)

jeanparpaillon:
Closes #256


whatyouhide:
Yep, sweet and easy! Thanks @jeanparpaillon 💟

PR #261, opened 2022-01-17
Update Protobuf.FieldProps

zolakeith:
This commit changes the typespec for FieldProps.type to include the
enumerated type.


whatyouhide:
Thanks @zolakeith! 💟

PR #264, opened 2022-01-19
Force prod env for escript.build task

jeanparpaillon:
When building protoc-gen-elixir plugin, we don't want to include
test/dev dependencies (stream_data, etc)


whatyouhide:
Thank you @jeanparpaillon! ❤️

PR #270, opened 2022-04-22
fix(new function): new/3 and new!/3 functions works with scalar types as previously

T0ha:
Fixed because team asked to get it back


T0ha:
yes, sorry

Issue #287, opened 2022-06-23
Dialyzer does not accept structs created with `new/0`, `new/1`... 


danmarcab:
If we have the following code:

```elixir
  alias Google.Protobuf.Timestamp

  @spec fine :: Timestamp.t()
  def fine do
    %Timestamp{seconds: 232_342, nanos: 232_342}
  end

  @spec not_fine :: Timestamp.t()
  def not_fine do
    Timestamp.new(seconds: 232_342, nanos: 232_342)
  end
```

The first one is ok according to dialyzer. The second one isn't. 
It seems the type spec of the behaviour https://github.com/elixir-protobuf/protobuf/blob/main/lib/protobuf.ex#L81 is `struct()` and dialyzer doesn't accept any struct as the concrete type.

I created a repo with a minimal case: 
https://github.com/danmarcab/protobuf_dialyzer_error_test_repo

danmarcab:
Sure, for the above code you get:

```
Total errors: 1, Skipped: 0, Unnecessary Skips: 0
done in 0m0.9s
lib/protobuf_dialyzer_error_test_repo.ex:26:invalid_contract
The @spec for the function does not match the success typing of the function.

Function:
ProtobufDialyzerErrorTestRepo.not_fine/0

Success typing:
@spec not_fine() :: %{:__struct__ => atom()}

________________________________________________________________________________
done (warnings were emitted)
Halting VM with exit status 2
```

whatyouhide:
Closed via #288.

PR #288, opened 2022-06-24
Add optional keys to Builder typespecs

danmarcab:
Related issue: https://github.com/elixir-protobuf/protobuf/issues/287

The previous typespecs weren't allowing types like:

```
  @spec not_fine :: Timestamp.t()
  def not_fine do
    Timestamp.new(seconds: 232_342, nanos: 232_342)
  end
```

Because the `Builder.new` typespecs were requiring a map with only the `:__struct__` key.

### DIFF
--- a/lib/protobuf/builder.ex
+++ b/lib/protobuf/builder.ex
@@ -3,17 +3,20 @@ defmodule Protobuf.Builder do
 
   alias Protobuf.FieldProps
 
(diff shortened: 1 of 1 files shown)


whatyouhide:
Good catch, thanks @danmarcab 💟

danmarcab:
Thank you and the rest of the team for the awesome work! 👍

PR #293, opened 2022-07-03
Add nicer oneof error messages

Cytro54:
When using `oneof`'s in client stubs, an user can specify a non-existing field. 
For example:
Imagine we have this proto :
```proto
message LoginRequest {
        reserved 420, 69;
        string username = 1;
        oneof credential {
            string password = 2;
            string googleOauthCode = 3;
            string magicLinkToken = 4;
       }
}
```
And the user writes this code : 
```elixir
req = Service.LoginRequest.new(username: "foo", credential: {:undefined_credential, "stuff"}) #oops
```
If he tries to encode this message, he'll be met with an unfriendly error message : `(MatchError) no match of right hand side value: nil`.

This PR implements a more friendly error message in these cases.

### DIFF
--- a/lib/protobuf/encoder.ex
+++ b/lib/protobuf/encoder.ex
@@ -208,7 +208,8 @@ defmodule Protobuf.Encoder do
 
               nil ->
                 raise Protobuf.EncodeError,
-                  message: "#{inspect(field_name)} wasn't found in #{inspect(struct.__struct__)}##{field}"
+                  message:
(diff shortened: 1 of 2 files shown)


whatyouhide (commented):
Can we add a test for this? 🙃

Cytro54:
sorry, forgot to add them :(

whatyouhide:
Last thing: looks like CI is yelling at us because of formatting 😄 Can you run `mix format` and push?

whatyouhide:
Thank you @Cytro54! 💟

Cytro54:
thx :)

PR #296, opened 2022-07-26
Fix use Bitwise warning

wingyplus:


whatyouhide:
Thank you @wingyplus! 💟

PR #302, opened 2022-08-17
Fix decoding of proto3 optional zero/default values

sindrip:
Currently with proto3 optional values, a zero/default value is skipped. This means that `decode(encode(msg))` is not identity for all messages with proto3 optional fields.

This PR ensures that only nil values for proto3_optional fields are skipped.


ericmj (approved):

whatyouhide:
Thanks @sindrip! 💟

PR #309, opened 2022-08-19
Fix conformance tests in CI

whatyouhide:
Protobuf got rid of autotools. See also https://github.com/protocolbuffers/protobuf/blob/main/cmake/README.md.

See also what they did [in Swift](https://github.com/apple/swift-protobuf/commit/29f4c00bb6bdbb5aa53f7c7ef38ce2782885d7c4).a


ericmj (approved):
💜

PR #325, opened 2022-12-12
Fix map fields contained in .proto files without a package

### LINKED ISSUE #321: How do I compile map fields to "map: true" instead of "repeated: true"?
Right now i edit the generated code by hand. Thank you!

rbino:
In `nested_maps/2`, the fully qualified name was built without discarding the nil entries of the list. This meant that map fields worked correctly for .proto files that contained a package, but failed to do so in .proto files that did not, since the map key had then an extra `.` in front that made the lookup in `get_field/4` fail.

This also adds a test case (that was failing with the old version and is now succeeding) to avoid regressions.

Fix #321

Signed-off-by: Riccardo Binetti <rbino@gmx.com>


whatyouhide:
Fantastic work @rbino, thanks! 💟

PR #329, opened 2022-12-14
Remove pack/1 and unpack/2 from Protobuf.Any

whatyouhide:
The idea here is to push people to use `Google.Protobuf.Any` directly more. The default `type_url` can be written down manually, and it's probably going to be a lot more accurate.

I'm thinking of removing `Protobuf.Any.type_url_to_module/1` too and make it a private function in this library instead of part of the public API. Thoughts @ericmj?

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,8 +1,8 @@
+locals_without_parens = [field: 2, field: 3, oneof: 2, extend: 4, extensions: 1]
+
 [
   inputs: ["{mix,.formatter}.exs", "{config,lib,conformance,test}/**/*.{ex,exs}"],
-  locals_without_parens: [field: 2, field: 3, oneof: 2, extend: 4, extensions: 1],
-  export: [
-    locals_without_parens: [field: 2, field: 3, oneof: 2, extend: 4, extensions: 1]
-  ],
+  locals_without_parens: locals_without_parens,
+  export: [locals_without_parens: locals_without_parens],
   import_deps: [:stream_data]
 ]
--- a/lib/elixirpb.pb.ex
+++ b/lib/elixirpb.pb.ex
@@ -3,8 +3,6 @@ defmodule Elixirpb.FileOptions do
 
   use Protobuf, protoc_gen_elixir_version: "0.11.0", syntax: :proto2
 
-  def fully_qualified_name, do: "elixirpb.FileOptions"
-
   field :module_prefix, 1, optional: true, type: :string
 end
 
--- a/lib/google/protobuf/compiler/plugin.pb.ex
+++ b/lib/google/protobuf/compiler/plugin.pb.ex
@@ -3,8 +3,6 @@ defmodule Google.Protobuf.Compiler.CodeGeneratorResponse.Feature do
 
   use Protobuf, enum: true, protoc_gen_elixir_version: "0.11.0", syntax: :proto2
 
-  def fully_qualified_name, do: "google.protobuf.compiler.CodeGeneratorResponse.Feature"
-
   field :FEATURE_NONE, 0
   field :FEATURE_PROTO3_OPTIONAL, 1
 end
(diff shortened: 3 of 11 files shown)


ericmj (approved):

yordis:
Could the `fully_qualified_name` rollback? This is causing some headaches, manual mapping is a lot of work compared to what other ecosystem

whatyouhide:
@yordis can you expand on the kind of manual mapping you have to do and why it's a lot of work/a headache?

yordis:
We need to extract the Full Name in order to use the value as the event type in the event store events table, and in the message bus we would use `google.protobuf.Any`.

Without "Full Name" function, this requires a lot of manual entry of information (unless I am doing something wrong, I am all ears), while in other programming languages that is taken care of (we are polyglot btw, Elixir, Go, TS at the moment)


```proto
syntax = "proto3";
package messagebus.v1;

import "google/protobuf/any.proto";
import "google/protobuf/timestamp.proto";

message Command {
  string id = 1;
  google.protobuf.Any data = 3;
  map<string, string> metadata = 4;
  google.protobuf.Timestamp created_at = 5;
  string correlation_id = 6;
  string causation_id = 7;
}
```

```proto
syntax = "proto3";
package myapp.v1;

message MyCommand {
  string command_id = 1;
}

message MyEventName {
  string event_id = 1;
}

message Event {
  oneof event {
    MyEventName my_event_name = 1;
  }
}
```


```go
func eventTypeProvider(event *eventv1.Event) (*string, error) {
	switch e := event.Event.(type) {
	case *eventv1.Event_MyEventName:
                // myapp.v1.MyEventName
		return MessageFullName(e.MyEventName)
	default:
		return nil, trogon.ErrUnknownEvent
	}
}


func MessageFullName[Message protoreflect.ProtoMessage](msg Message) string {
	return msg.ProtoReflect().Descriptor().FullName()
}
```


```go
func TestExample(t *testing.T) {
	anymsg, err := anypb.New(&myappv1.MyCommand{
		CommandId: "123",
	})
	require.NoError(t, err)

	message := messagebus.CommandMessage{Data: anymsg}

	marshal, err := protojson.Marshal(&message)
	require.NoError(t, err)
	fmt.Sprintf("%v", string(marshal))
        // {"data":{"@type":"type.googleapis.com/myapp.v1.MyCommand", "commandId":"123"}}

	eventype, err := eventTypeProvider(&myappv1.Event{
		Event: &myappv1.Event_MyEventName{
			MyEventName: &myappv1.MyEventName{
				EventId: "123",
			},
		},
	})
	require.NoError(t, err)
	fmt.Sprintf("%v", eventype)
        // myapp.v1.MyEventName
}

func eventTypeProvider(event *myappv1.Event) (string, error) {
	switch e := event.Event.(type) {
	case *myappv1.Event_MyEventName:
		return MessageFullName(e.MyEventName), nil
	default:
		return "", ErrUnknownEvent
	}
}

func MessageFullName[Message protoreflect.ProtoMessage](msg Message) string {
	return string(msg.ProtoReflect().Descriptor().FullName())
}
```


https://pkg.go.dev/google.golang.org/protobuf/types/known/anypb#MarshalFrom
```go
// MarshalFrom marshals src into dst as the underlying message
// using the provided marshal options.
//
// If no options are specified, call dst.MarshalFrom instead.
func MarshalFrom(dst *Any, src proto.Message, opts proto.MarshalOptions) error {
	const urlPrefix = "type.googleapis.com/"
	if src == nil {
		return protoimpl.X.NewError("invalid nil source message")
	}
	b, err := opts.Marshal(src)
	if err != nil {
		return err
	}
	dst.TypeUrl = urlPrefix + string(src.ProtoReflect().Descriptor().FullName())
	dst.Value = b
	return nil
}
```

whatyouhide:
Is `full_name/0` able to predictably return, 100% of the time, the correct full name? I completely lost context on this change, so I’m trying to understand why in the PR description I say constructing URLs manually would be "a lot more accurate".

yordis:
> Is full_name/0 able to predictably return, 100% of the time, the correct full name?

That is what other SDKs are doing, and everything works around such value when dealing with Any, so it is expected to be deterministic value.

> "a lot more accurate".

It confused me as well, I was trying to find information about what was the precise problem as well to make sure I understood the problem, I am coming from the perspective that other SDKs do not have such issue, as far as I can tell

PR #330, opened 2022-12-18
Deprecate the `new(!)/0,1` callbacks

whatyouhide:
We now fully support struct building with compile-time defaults, so I think it's time to deprecate these and reduce the API surface.

This is already a huge library, and simplifying it is paramount (IMO) to keeping its maintenance sustainable.

The change from `MyMessage.new(attrs)` to `%MyMessage{attrs}` is (tedious but) straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ end
 
       field :name, 1, type: :string
     end
-    
+
     defmodule Helloworld.HelloReply do
       @moduledoc false
       use Protobuf, protoc_gen_elixir_version: "0.10.0", syntax: :proto3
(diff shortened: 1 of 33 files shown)


ericmj (approved):

hamir-suspect:
Hello, I was wondering if this change had some other goal than reducing API surface, as I am not involved with the project so far I can't critique the choices but I'd like to offer my opinion on this and a usecase that I think goes "elixir way"
in some codebases `MyMessage.new/1` was also used to pipe attrs into as:
`attrs |> MyMessage.new()` 
so this change removed a nice convinient way to build structs.

PR #334, opened 2023-01-03
Update docs to work with v0.10.0

D-Town44:
Old docs in `Protobuf` module (and hex docs) don't work with a 0.10.0 change not requiring `defstruct` when making a new protobuf module.

Before change:

image

After change:

image

Because of this change, the example from the Protobuf behaviour hex docs can be used without edits.


whatyouhide:
Thanks @D-Town44! 💟

PR #336, opened 2023-01-10
Add support for handling "." in FieldMask paths when JSON decoding

### LINKED ISSUE #335: JSON decoding should work with "." in FieldMask path
Currently JSON decoding throws a `:bad_field_mask` error if a `FieldMask` path contains a period. According to the [official docs](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#json-encoding-of-field-masks), a period is a valid separator, used to denote nested paths. e.g. `user.display_name`. FWIW, the encoding side seems to be fine..

```elixir
{:error,
    %Protobuf.JSON.DecodeError{
        message: "invalid characters in field mask: \"user.display_name\""
    }}
```

jareddellitt:
Fixes #335


whatyouhide:
Great catch, thanks @jareddellitt! 💟

PR #352, opened 2023-06-14
Add protobuf code comments as Elixir module documentation

btkostner:
This PR takes code comments from the protobuf files and add them as `@moduledoc` blocks. Some of the code and all of the magic numbers are taken from [protokit golang library](https://github.com/pseudomuto/protokit/blob/master/parser.go). This is the same parser logic used by https://github.com/pseudomuto/protoc-gen-doc which is pretty popular.

Example generated code:

```elixir
defmodule My.Test.Days do
  @moduledoc """
  This enum represents days of the week.
  """

  use Protobuf, enum: true, protoc_gen_elixir_version: "0.12.0", syntax: :proto2

  field :MONDAY, 1
  field :TUESDAY, 2
  field :LUNDI, 1
end
```

Todos:
- [x] Make the code cleaner. The recursive index based path for comments is pretty annoying to deal with.
- [x] Get documentation for messages working
- [x] Get documentation for services working
- [ ] ~Documentation for fields?~ (This does nothing, but would be nice for future improvements)
- [x] Tests


btkostner:
Alright, this should be a good starting point. It generates the code comments as expected. Code is pretty cleaned up. I think adding some work for documenting individual fields will be the next step, but we should be good here so far.

lib/protobuf/protoc/generator.ex
```diff
@@ -36,13 +36,20 @@ defmodule Protobuf.Protoc.Generator do
...
           dep_type_mapping: get_dep_type_mapping(ctx, desc.dependency, desc.name)
       }
       |> Protobuf.Protoc.Context.custom_file_options_from_file_desc(desc)
 
-    enum_defmodules = Enum.map(desc.enum_type, &Generator.Enum.generate(ctx, &1))
+    enum_defmodules =
+      desc.enum_type
+      |> Enum.with_index()
+      |> Enum.map(fn {enum, index} ->
+        {Context.append_comment_path(ctx, "5.#{index}"), enum}
```
Nezteb, line 50: What is the hard-coded `5.` for specifically?

lib/protobuf/protoc/generator.ex
```diff
@@ -51,7 +58,14 @@ defmodule Protobuf.Protoc.Generator do
 
     service_defmodules =
       if "grpc" in ctx.plugins do
-        Enum.map(desc.service, &Generator.Service.generate(ctx, &1))
+        desc.service
+        |> Enum.with_index()
+        |> Enum.map(fn {service, index} ->
+          Generator.Service.generate(
+            Context.append_comment_path(ctx, "6.#{index}"),
```
Nezteb, line 65: Same question for the hard-coded `6.`. 😄

Nezteb:
@btkostner Regarding your comment:
>  ~~Documentation for fields?~~ (This does nothing, but would be nice for future improvements)

Could you elaborate on the "this does nothing" part? 😄

btkostner:
> @btkostner Regarding your comment:
> 
> > ~Documentation for fields?~ (This does nothing, but would be nice for future improvements)
> 
> Could you elaborate on the "this does nothing" part? 😄

Two ways to document the fields. Add a `doc` option for fields which I think would be preferred. Otherwise, just create a `@typedoc` block with each field documented. I decided not to implement the field documentation until the team decides on the way forward.

lib/protobuf/protoc/generator.ex
btkostner: This number is from FileDescriptorProto. Copied from [the upstream code](https://github.com/pseudomuto/protokit/blob/dc6fd1c8003a61bcd77ddc148e4d49946fbf8c23/parser.go#L13).

lib/protobuf/protoc/generator.ex
btkostner: This number is from FileDescriptorProto. Copied from [the upstream code](https://github.com/pseudomuto/protokit/blob/dc6fd1c8003a61bcd77ddc148e4d49946fbf8c23/parser.go#L13).

Nezteb:
> until the team decides on the way forward

@whatyouhide @wingyplus Y'all show up often in the git history of the files Blake modified, so I feel like you are the most qualified to chime in. Do either of you have a preference on field docs?

I'd add you as PR reviewers but I don't have permissions. 😅

mix.exs
```diff
@@ -49,6 +49,7 @@ defmodule Protobuf.Mixfile do
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
       {:credo, "~> 1.5", only: [:dev, :test], runtime: false},
       {:ex_doc, ">= 0.0.0", only: :dev, runtime: false},
+      {:grpc, "~> 0.6", only: :test},
```
v0idpwn, line 52: I think `:protobuf` can't depend on `:grpc`, because `:grpc` depends on `:protobuf`. This is a somewhat recent development, at `:grpc` 0.6, it did not.

btkostner: Mhmm, this makes it hard to test the grpc codegen (I think, it's been a while since I've looked at this code.) I'll have to think about some other way to make this work nicely.

v0idpwn: I think an option is to create a separate package to test this. It doesn't look great, but does the job. I'd love to merge this and will try to invest some time to play with your branch over the next weeks (I'm slow, got a lot of work).

btkostner: Noted. I'll try to get to this, though my week is also pretty busy so I can't give a timeline.

v0idpwn: Hi, @btkostner! 
I just pushed [a branch](https://github.com/elixir-protobuf/protobuf/tree/documentation-2) removing the dependency. I think the test definitely tests less than the one you have here, but I think the parsing functionality is already covered elsewhere, so it's probably fine to do it like I did. WDYT? Should I push these commits into your branch?

btkostner: Code looks good. Push away!

priv/templates/service.ex.eex
```diff
@@ -1,7 +1,12 @@
 defmodule <%= @module %>.Service do
-  <%= unless @module_doc? do %>
+  <%= if @comment != "" do %>
+  @moduledoc """
+  <%= @comment %>
+  """
+  <% else %>
   @moduledoc false
   <% end %>
```
v0idpwn, line 8: I disagree a bit here. I don't think we should `@moduledoc false` if there's no comments for a file and `include_docs` is true. For example, I have an application where we use `include_docs=true` because we want the typespecs and structs to be public in the generated documentation.

v0idpwn: Pushed this too

v0idpwn:
@whatyouhide @ericmj I'd really like to merge this. The only potential blocker I see are warnings of this type for Elixir 1.12 / OTP 24:

```
The current heredoc line is indented too little
  nofile:6

warning: outdented heredoc line. The contents inside the heredoc should be indented at the same level as the closing """. The following is forbidden:

    def text do
      """
    contents
      """
    end

Instead make sure the contents are indented as much as the heredoc closing:

    def text do
      """
      contents
      """
    end
```

v0idpwn:
Well, I'm merging :)
If there's a post-merge review, I can fix any comments.

Thank you, @btkostner! 
Over the next weeks, I'll try to get to your Google protos PR too.

PR #353, opened 2023-06-17
Generate Google modules

btkostner:
This includes the well known Google protobuf modules to deprecate the `google_protos` package. From this comment: https://github.com/elixir-protobuf/google-protos/pull/21#issuecomment-1595737724.

PR #354, opened 2023-06-17
Add helper functions for Google.Protobuf modules

btkostner:
Replaces https://github.com/elixir-protobuf/google-protos/pull/21. Should be merged after https://github.com/elixir-protobuf/protobuf/pull/353.


Nezteb (approved):

ddengler (approved):

lib/google/protobuf.ex
```diff
@@ -0,0 +1,186 @@
...
+  ## Examples
+
+      iex> to_time_unit(%Google.Protobuf.Duration{seconds: 10})
+      {10, :second}
+
+      iex> to_time_unit(%Google.Protobuf.Duration{seconds: 20, nanos: 100})
+      {20_000_000_100, :nanosecond}
+
+  """
+  @spec to_time_unit(Google.Protobuf.Duration.t()) :: {integer(), System.time_unit()}
```
btkostner, line 18: Heads up, now that we have https://hexdocs.pm/elixir/main/Duration.html, we should change these functions to match.

v0idpwn: We would need to bump elixir requirement, though

v0idpwn: I think it's fine to keep as it is?

v0idpwn: Maybe we can **accept** it in `from_time_unit` if using Elixir >= 1.17, but continue returning ` {integer(), System.time_unit()}` on `to_time_unit`

v0idpwn (approved):

lib/google/protobuf.ex
whatyouhide: If we don't return `Duration`, we're introducing a function now that is bound to be deprecated soon. I think I wouldn't define `to_time_unit/1` in the first place for now, or only defined it if compiled with Elixir 1.17+.

whatyouhide: Same for `from_time_unit`

lib/google/protobuf.ex
```diff
@@ -0,0 +1,186 @@
...
+
+  def from_time_unit(millisecond, :nanosecond) do
+    struct(Google.Protobuf.Duration, %{
+      seconds: div(millisecond, 1_000_000_000),
+      nanos: rem(millisecond, 1_000_000_000)
+    })
+  end
+
+  @doc """
+  Converts a `Google.Protobuf.Struct` struct to a `map()` recursively
```
whatyouhide, line 68: Please use the correct ExDoc syntax for linking types:

```suggestion
  Converts a `Google.Protobuf.Struct` struct to a `t:map/0` recursively
```

v0idpwn:
@btkostner, would you like to address Andrea's comments, or would you prefer for me to take this one over? Happy either way :)

btkostner:
Go for it! I'm on a deadline for work right now so I doubt I'll have free time to work on this 👍

lib/google/protobuf.ex
v0idpwn: Removed them for now, we can consider adding them at some point :)

yordis:
@v0idpwn do you think it would be possible to cut off a release?

v0idpwn:
Hi, @yordis! I don't have the permission to make releases, so we need to ask @whatyouhide or @ericmj. 

For me, the current main seems okay to be released. I'm just a bit worried that #384 and #353 may make update not very smooth.

yordis:
Removing the google_protos pkg will fix #353 since I just did it. Thus far, things are working for me; what is your concern with #384 per se?

PR #358, opened 2023-08-29
Add optional keys to `Protobuf.decode/2` typespecs

danmarcab:
Same issue as in https://github.com/elixir-protobuf/protobuf/issues/287, the typespec was over constrained and not really accepting the structs as it should

### DIFF
--- a/lib/protobuf.ex
+++ b/lib/protobuf.ex
@@ -175,7 +175,11 @@ defmodule Protobuf do
       #=> ** (Protobuf.DecodeError) ...
 
   """
-  @spec decode(binary(), message) :: %{required(:__struct__) => message} when message: module()
+  @spec decode(binary(), message) :: %{
+          required(:__struct__) => message,
+          optional(atom()) => any()
+        }
+        when message: module()
   defdelegate decode(data, module), to: Protobuf.Decoder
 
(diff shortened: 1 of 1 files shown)


lib/protobuf.ex
```diff
@@ -175,7 +175,11 @@ defmodule Protobuf do
       #=> ** (Protobuf.DecodeError) ...
 
   """
-  @spec decode(binary(), message) :: %{required(:__struct__) => message} when message: module()
+  @spec decode(binary(), message) :: %{
+          required(:__struct__) => message,
+          optional(atom()) => any()
+        }
+        when message: module()
```
whatyouhide, line 182: I think we can change this to

```suggestion
  @spec decode(binary(), message) :: struct()
```

?

danmarcab: I guess that way we'd lose the fact that the struct is not any struct but one of the `message` type.
Will fix our problem, so we can go with that if you prefer

whatyouhide: Aaaah, I missed that, sorry. Okay no, your fix is great. 🎉

whatyouhide:
Thank you @danmarcab! 💟

PR #359, opened 2023-09-19
Fix JSON encoding of proto3 optional non-nil defaults

zookzook:
It fixes the same issue but now for the JSON encoder like https://github.com/elixir-protobuf/protobuf/pull/302

Currently with proto3 optional values, a zero/default value is skipped. This means that the JSON `decode(encode(msg))` is not the same for all messages with proto3 optional fields.

This PR ensures that attributes containing the default values are not skipped for proto3_optional fields.

### DIFF
--- a/lib/protobuf/json/encode.ex
+++ b/lib/protobuf/json/encode.ex
@@ -162,8 +162,22 @@ defmodule Protobuf.JSON.Encode do
   defp encode_regular_fields(struct, %{field_props: field_props}, opts) do
     for {_field_num, %{name_atom: name, oneof: nil} = prop} <- field_props,
         %{^name => value} = struct,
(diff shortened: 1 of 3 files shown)


lib/protobuf/json/encode.ex
```diff
@@ -162,11 +162,25 @@ defmodule Protobuf.JSON.Encode do
...
+  end
+
+  defp emit?(prop, value) do
+    case default?(prop, value) do
+      true ->
+        prop.proto3_optional?
+
+      false ->
+        true
+    end
```
whatyouhide, line 181:
```suggestion
    if default?(prop, value), do: prop.proto3_optional?, else: true
```

whatyouhide:
Thanks @zookzook 💟

PR #377, opened 2024-09-25
Keep same alphabetical order in `Util.options_to_str/1`

### LINKED ISSUE #361: Consideration in generating options in Messages in OTP 26
Considering that options are [map](https://github.com/elixir-protobuf/protobuf/blob/6c580a2a515c6a1f7a6ec96ecaeed6e0d215dc61/lib/protobuf/protoc/generator/message.ex#L91-L96) and in OTP 26 the order keys of map [is not granted](https://www.erlang.org/blog/otp-26-highlights/#changed-ordering-of-atom-keys), should change it to keyword list?

alternatively [Util.options_to_str](https://github.com/elixir-protobuf/protobuf/blob/main/lib/protobuf/protoc/generator/util.ex#L47) and friends can run sort by key before join map_join?

ianko:
Fixes #361.

---

Solves the issue where the module options keep changing on OTP-26+.

### DIFF
--- a/lib/protobuf/protoc/generator/util.ex
+++ b/lib/protobuf/protoc/generator/util.ex
@@ -47,6 +47,7 @@ defmodule Protobuf.Protoc.Generator.Util do
   def options_to_str(opts) when is_map(opts) do
     opts
     |> Enum.reject(fn {_key, val} -> val in [nil, false] end)
+    |> Enum.sort()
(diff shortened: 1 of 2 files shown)


hugo-tubi (approved):

v0idpwn (approved):

whatyouhide:
Thanks @ianko 💟

PR #379, opened 2024-10-20
Keep maps (actually, structs) for oneof

### LINKED ISSUE #362: Inconsistent behaviour when encoding oneof
Hi there,

We're seeing (what feels like) inconsistent behaviour when encoding oneof value.

Given the following protobuf:

``` protobuf
syntax = "proto3";

message Foo {
  message Bar {
    string value = 1;
  }

  message Baz {}

  oneof value {
    Bar bar = 1;
    Baz baz = 2;
  }
}
```

Which compiles to the following elixir code:

```elixir
defmodule Foo.Bar do
  @moduledoc false

  use Protobuf, protoc_gen_elixir_version: "0.12.0", syntax: :proto3

  field(:value, 1, type: :string)
end

defmodule Foo.Baz do
  @moduledoc false

  use Protobuf, protoc_gen_elixir_version: "0.12.0", syntax: :proto3
end

defmodule Foo do
  @moduledoc false

  use Protobuf, protoc_gen_elixir_version: "0.12.0", syntax: :proto3

  oneof(:value, 0)

  field(:bar, 1, type: Foo.Bar, oneof: 0)
  field(:baz, 2, type: Foo.Baz, oneof: 0)
end
```

When encoding/decoding a message that contains a map for `Foo.Bar` (not the struct) and it contains a value - things work as expected:

```elixir
foo = %Foo{value: {:bar, %{value: "bar"}}}
foo |> Foo.encode() |> Foo.decode()

%Foo{
  value: {:bar, %Foo.Bar{value: "bar", __unknown_fields__: []}},
  __unknown_fields__: []
}
```

But when encoding/decoding an empty map - it simply gets dropped entirely - even though we are passing the proper tuple for the oneof:

```elixir
foo = %Foo{value: {:baz, %{}}}
foo |> Foo.encode() |> Foo.decode()

%Foo{value: nil, __unknown_fields__: []}
```

Since we've noted which oneof it is -  it seems like it should be included.

Thanks

### LINKED ISSUE #365: `Protobuf.encode/1` does not actually verify struct in some cases

v0idpwn:
Closes #362. 

We are consistently accepting/coercing maps into structs, so I don't see a problem in doing what I did. Maps themselves aren't accepted in oneofs, so no impact on them.

### DIFF
--- a/lib/protobuf/encoder.ex
+++ b/lib/protobuf/encoder.ex
@@ -73,19 +73,18 @@ defmodule Protobuf.Encoder do
           "Got error when encoding #{inspect(struct.__struct__)}##{prop.name_atom}: #{Exception.format(:error, error)}"
   end
 
+  defp skip_field?(_syntax, val, %FieldProps{oneof: oneof}) when not is_nil(oneof), do: is_nil(val)
   defp skip_field?(_syntax, [], _prop), do: true
   defp skip_field?(_syntax, val, _prop) when is_map(val), do: map_size(val) == 0
   defp skip_field?(:proto2, nil, %FieldProps{optional?: optional?}), do: optional?
   defp skip_field?(:proto2, value, %FieldProps{default: value, oneof: nil}), do: true
 
-  defp skip_field?(:proto3, val, %FieldProps{proto3_optional?: true}),
-    do: is_nil(val)
-
+  defp skip_field?(:proto3, val, %FieldProps{proto3_optional?: true}), do: is_nil(val)
   defp skip_field?(:proto3, nil, _prop), do: true
-  defp skip_field?(:proto3, 0, %FieldProps{oneof: nil}), do: true
-  defp skip_field?(:proto3, +0.0, %FieldProps{oneof: nil}), do: true
-  defp skip_field?(:proto3, "", %FieldProps{oneof: nil}), do: true
-  defp skip_field?(:proto3, false, %FieldProps{oneof: nil}), do: true
+  defp skip_field?(:proto3, 0, _prop), do: true
+  defp skip_field?(:proto3, +0.0, _prop), do: true
+  defp skip_field?(:proto3, "", _prop), do: true
+  defp skip_field?(:proto3, false, _prop), do: true
   defp skip_field?(_syntax, _val, _prop), do: false
 
   defp encode_field(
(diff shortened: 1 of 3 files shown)


v0idpwn:
Closing this in favor of https://github.com/elixir-protobuf/protobuf/pull/398/commits/b5258459ca9f8f761688dd78cace76ea954b4446 which does the same with a better approach

PR #384, opened 2024-12-04
Add `typespec/1` macro to transform module behaviour

### LINKED ISSUE #382: Types using protobuf transformer have wrong typespecs
When a protobuf module implements `transform_module/0`, the autogenerated typespecs usually become invalid. 

Assume the following example (extracted from `Protobuf.TransformModule` docs):

```elixir
defmodule StringMessage do
  use Protobuf, syntax: :proto3

  field :value, 1, type: :string

  def transform_module(), do: MyTransformModule
end
```

With the following transformer (also extracted):
```elixir
defmodule MyTransformModule do
  @behaviour Protobuf.TransformModule

  @impl true
  def encode(string, StringMessage) when is_binary(string), do: %StringMessage{value: string}

  @impl true
  def decode(%StringMessage{value: string}, StringMessage), do: string
end
```

You can use the above type as a string:
```
iex(4)> defmodule Foo do
...(4)>     use Protobuf, syntax: :proto3
...(4)>     field :string, 1, type: StringMessage
...(4)> end
iex(5)> Protobuf.encode(%Foo{string: "hello"})
<<10, 7, 10, 5, 104, 101, 108, 108, 111>>
iex(6)> Protobuf.decode(v, Foo)
%Foo{string: "hello", __unknown_fields__: []}
```

If you try to use a struct, it fails:
```
iex(7)> Protobuf.encode(%Foo{string: %StringMessage{value: "hello"}})
** (Protobuf.EncodeError) Got error when encoding Foo#string: ** (FunctionClauseError) no function clause matching in MyTransformModule.encode/2
```

Yet the auto-generated typespec doesn't reflect that. Instead, the typespec will be:
```elixir
@type t() :: %__MODULE__{value: String.t()}
```

This causes valid code to be flagged by dialyzer as invalid, and invalid code to be considered valid.

v0idpwn:
Adds a `typespec/1` macrocallback to transform module behaviour.

This macro can modify the typespec for modules using it. This way, we can provide correct typespecs for modules using transformers. A "breaking" change is that transform modules can't depend on Protobuf definitions anymore. This is quite easily solvable, though.

Closes #382 

--------------------------------------------------
**Previous description, from previous implementation:**

We detect that `transform_module/0` was defined through a on_definition hook. We then store the module in an attribute. This attribute is later used by the `before_compile` callback to set the typespec, which is set in the form of:

```elixir
@type t() :: transform_module.t(__MODULE__)
```

Then, transform modules can implement proper typespecs.

No changes if `transform_module` is not overriden.

Closes #382

### DIFF
--- a/lib/protobuf/dsl.ex
+++ b/lib/protobuf/dsl.ex
@@ -138,11 +138,11 @@ defmodule Protobuf.DSL do
 
   # Registered as the @on_definition compile callback for modules that call "use Protobuf"
   # Allow us to detect when `transform_module` is re-defined
-  def on_def(_env, :def, :transform_module, [], [], [do: nil]) do
+  def on_def(_env, :def, :transform_module, [], [], do: nil) do
     :ok
   end
 
-  def on_def(env, :def, :transform_module, [], [], [do: module_alias_ast]) do
+  def on_def(env, :def, :transform_module, [], [], do: module_alias_ast) do
     Module.put_attribute(env.module, :transform_module, module_alias_ast)
     :ok
   end
@@ -242,9 +242,34 @@ defmodule Protobuf.DSL do
     end
   end
 
-  defp def_t_typespec(_props, _extension_props, transform_module_ast) when not is_nil(transform_module_ast) do
+  defp def_t_typespec(props, extension_props, transform_module_ast)
+       when not is_nil(transform_module_ast) do
+    default_typespec = def_t_typespec(props, extension_props, nil)
+
     quote do
-      @type t() :: unquote(transform_module_ast).t(__MODULE__)
+      case Code.ensure_compiled(unquote(transform_module_ast)) do
+        {:module, _} ->
+          :ok
+
+        _ ->
+          raise CompileError, """
+          Transform module #{inspect(unquote(transform_module_ast))} not available
+          during protobuf definition compilation.
+
+          Since protobuf v0.14, protobuf definitions depend in compile time on
+          their transform modules. This means that transform modules can't depend
+          on protobuf structs, and must be available for compilation when protobuf
+          definitions are compiled.
+          """
+      end
+
+      require unquote(transform_module_ast)
+
+      if macro_exported?(unquote(transform_module_ast), :typespec, 2) do
+        unquote(transform_module_ast).typespec(__MODULE__, unquote(default_typespec))
+      else
+        unquote(default_typespec)
+      end
     end
   end
 
--- a/lib/protobuf/dsl.ex
+++ b/lib/protobuf/dsl.ex
@@ -254,11 +254,11 @@ defmodule Protobuf.DSL do
         _ ->
           raise CompileError, """
           Transform module #{inspect(unquote(transform_module_ast))} not available
-          during protobuf definition compilation.
+          during Protobuf definition compilation.
 
-          Since protobuf v0.14, protobuf definitions depend in compile time on
+          Since Protobuf v0.14, Protobuf definitions depend in compile time on
           their transform modules. This means that transform modules can't depend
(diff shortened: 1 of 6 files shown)


v0idpwn:
@whatyouhide @ericmj if you agree on approach here, I'll add a few tests, documentation, etc :)

lib/protobuf/dsl.ex
```diff
@@ -226,19 +242,25 @@ defmodule Protobuf.DSL do
     end
   end
 
-  defp def_t_typespec(%MessageProps{enum?: true} = props, _extension_props) do
+  defp def_t_typespec(_props, _extension_props, transform_module_ast) when not is_nil(transform_module_ast) do
+    quote do
+      @type t() :: unquote(transform_module_ast).t(__MODULE__)
```
v0idpwn, line 247: The biggest potential issue is that we are assuming that the function ast will always be a module. But I can't see it being something else if you are using the library properly :)

v0idpwn:
I tested this in our codebases and it resolves dialyzer issues. We are using transformers for, for example, well known types.

v0idpwn:
JFYI, I don't think the typespecs as they stand in this PR are good enough. I'm working on a new implementation that, although more complex, provides better typespecs.

lib/protobuf/dsl.ex
```diff
@@ -226,19 +242,50 @@ defmodule Protobuf.DSL do
...
+       when not is_nil(transform_module_ast) do
+    default_typespec = def_t_typespec(props, extension_props, nil)
+
+    quote do
+      case Code.ensure_compiled(unquote(transform_module_ast)) do
+        {:module, _} ->
+          :ok
+
+        _ ->
+          raise CompileError, """
```
v0idpwn, line 255: Need to add a test for this error

lib/protobuf/transform_module/infer_fields_from_enum.ex
```diff
@@ -5,6 +5,8 @@ defmodule Protobuf.TransformModule.InferFieldsFromEnum do
   `c:Protobuf.new/1` before the value is encoded.
   """
 
+  @type t(_module) :: term()
+
```
v0idpwn, line 9:
```suggestion
```

lib/protobuf/dsl.ex
```diff
@@ -226,19 +242,50 @@ defmodule Protobuf.DSL do
...
+          Transform module #{inspect(unquote(transform_module_ast))} not available
+          during protobuf definition compilation.
```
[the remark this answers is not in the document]
whatyouhide: ...and all instances below

lib/protobuf/transform_module.ex
```diff
@@ -50,4 +60,13 @@ defmodule Protobuf.TransformModule do
   Called after a message is decoded.
   """
   @callback decode(message(), type()) :: value()
+
+  @doc """
+  Transforms the typespec for modules using this transformer
```
whatyouhide, line 65:
```suggestion
  Transforms the typespec for modules using this transformer.
```

lib/protobuf/transform_module.ex
```diff
@@ -50,4 +60,13 @@ defmodule Protobuf.TransformModule do
...
+
+  Optional. If this callback is not present, the default typespec will be used.
```
whatyouhide, line 67: ExDoc already marks this as optional thanks to `@optional_callbacks`

```suggestion
  If this callback is not present, the default typespec will be used.
```

v0idpwn:
@whatyouhide I believe I addressed all your comments, plus I made a few simplifications. I'll go ahead and merge this :)

Issue #364, opened 2023-11-30
Can JSON decoder be loose on constraints about float? 


junaid1460:
Most JSON encoders just wrap 1.0 to 1, while protobuf gives good enough code generation, can we cast 1 -> 1.0 for floats and doubles?

The alternate solution is to use FloatValue and DoubleValue, that feels like just more work for no reason

whatyouhide:
@junaid1460 thanks for the report! Do you know what other Protobuf libraries (for other languages) do in these cases?

junaid1460:
We've been using protobuf in dart/flutter its json decoder just automatically handles 1 and spits out 1.0. 

eg: 
```js
// 1
{"value": 1} 

// 2
{"value": 2.3}

```

```proto
message Resp {
    float value;
}
```

when we parse `1` with `Resp` it should just take it as float.

whatyouhide:
Gotcha, what do decoders for Ruby and Python do for example?

v0idpwn:
Closed per #385

junaid1460:
Thanks :)

PR #385, opened 2024-12-05
Decode integer-formatted JSON numbers as float (#364)

warrenseine:
JSON has a single type for numbers. The same number can be written as `42` or `42.0`. Without additional context, this number should be valid in any of the Protobuf numeric types. For reference, the `protobuf` Python library is totally fine decoding `{ "foo": 42 }` with either a `float` or an `int32` Protobuf field.

### DIFF
--- a/lib/protobuf/json/decode.ex
+++ b/lib/protobuf/json/decode.ex
@@ -392,6 +392,7 @@ defmodule Protobuf.JSON.Decode do
   end
 
   defp decode_float(float) when is_float(float), do: {:ok, float}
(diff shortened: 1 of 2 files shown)


v0idpwn (approved):

v0idpwn:
Thank you! 
For the record, this is also specified in ProtoJSON documentation.

v0idpwn:
Hi, @warrenseine, I don't have the rights for doing releases. But it was already requested to Eric and Andrea, I believe it will be released soonish :)

warrenseine:
For anyone interested, this was fixed in 0.14.0 (thanks again!).

PR #387, opened 2024-12-28
Ensure there's a newline in the end of generated files

### LINKED ISSUE #380: Missing new line in the generate files
## Problem 

After the files are generated, despite the fact that we pass it to [Generator.Util.format/1](https://github.com/elixir-protobuf/protobuf/blob/18ad14f571cf296a1b87c9b5da559807da324be1/lib/protobuf/protoc/generator.ex#L27C14-L27C37) when they get saved into file the code is missing a `\n` char.

The problem arias because we expect all the source code in our repo pass `mix format` check but the formatter doesn't like the files are missing `\n` at the end of the file. We usually have to run `mix format` after files are generated which is fine but then on the next regenerate, all the files get changed.

## Expected

We add extra `\n` to all places we generate the code!

### Question

Where do you think is the right place to add newline? in `Generator.Util.format/1` or add it to all places that call `Generator.Util.format/1` ?

v0idpwn:
We must manually add it, as `Code.format_string!` doesn't add a newline to end of file, nor does protoc.

Closes #380

### DIFF
--- a/lib/protobuf/protoc/generator/util.ex
+++ b/lib/protobuf/protoc/generator/util.ex
@@ -81,11 +81,14 @@ defmodule Protobuf.Protoc.Generator.Util do
 
   @spec format(String.t()) :: String.t()
   def format(code) when is_binary(code) do
-    formatted_code = Code.format_string!(code, locals_without_parens: @locals_without_parens)
+    formatted_code =
+      code
+      |> Code.format_string!(locals_without_parens: @locals_without_parens)
+      |> IO.iodata_to_binary()
 
     # As neither Code.format_string!/2 nor protoc automatically adds a newline
-    # at end of files, we must add ourselves
-    IO.iodata_to_binary([formatted_code, "\n"])
+    # at end of files, we must add ourselves if not present.
+    if String.ends_with?(formatted_code, "\n"), do: formatted_code, else: formatted_code <> "\n"
   end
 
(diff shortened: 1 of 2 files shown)


lib/protobuf/protoc/generator/util.ex
```diff
@@ -81,9 +81,11 @@ defmodule Protobuf.Protoc.Generator.Util do
...
   @spec format(String.t()) :: String.t()
   def format(code) when is_binary(code) do
-    code
-    |> Code.format_string!(locals_without_parens: @locals_without_parens)
-    |> IO.iodata_to_binary()
+    formatted_code = Code.format_string!(code, locals_without_parens: @locals_without_parens)
+
+    # As neither Code.format_string!/2 nor protoc automatically adds a newline
+    # at end of files, we must add ourselves
+    IO.iodata_to_binary([formatted_code, "\n"])
```
whatyouhide, line 88:
```suggestion
    formatted_code =
      code
      |> Code.format_string!(locals_without_parens: @locals_without_parens)
      |> IO.iodata_to_binary()

    # As neither Code.format_string!/2 nor protoc automatically adds a newline
    # at end of files, we must add ourselves if not present.
    if String.ends_with?(formatted_code, "\n"), do: formatted_code, else: formatted_code <> "\n"
```

Maybe this will be more resilient in the future?

PR #390, opened 2024-12-29
Support json_name for proto2; Ensures proper field emission for JSON encoder

v0idpwn:
Fixes >500 conformance tests
- Makes proto2 also support json_name. This alone fixes >200 tests
- Ensures JSON encoder decides correctly when to emit or not emit fields (at the very least, more correctly than before :))


v0idpwn:
I suppose this is also not controversial at all, so merging. We can revert before release if there's any problem.

PR #391, opened 2024-12-30
Fix decoding of maps to match conformance tests

v0idpwn:
Different from the previous fixes in #390, I'm not 100% sure about this one. Feels like it could/should be solved in a different layer, but I didn't find a better way to do it without breaking anything else.

### DIFF
--- a/lib/protobuf/decoder.ex
+++ b/lib/protobuf/decoder.ex
@@ -234,7 +234,15 @@ defmodule Protobuf.Decoder do
       prop
 
     embedded_msg = decode(bin, type)
-    val = if map?, do: %{embedded_msg.key => embedded_msg.value}, else: embedded_msg
+
+    val =
+      if map?,
+        do: %{
+          (embedded_msg.key || map_default(prop, :key)) =>
+            embedded_msg.value || map_default(prop, :value)
+        },
+        else: embedded_msg
+
     val = if oneof, do: {name_atom, val}, else: val
 
     cond do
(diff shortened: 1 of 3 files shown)


test/protobuf/conformance_regressions_test.exs
```diff
@@ -83,6 +83,12 @@ defmodule Protobuf.ConformanceRegressionsTest do
 
     @describetag message_type: "protobuf_test_messages.proto3.TestAllTypesProto3"
 
+    test "map i32 i32" do
+      mod = ProtobufTestMessages.Proto2.TestAllTypesProto2
+      problematic_payload = <<194, 3, 0>>
+      assert m = %{map_int32_int32: %{0 => 0}} = mod.decode(problematic_payload)
```
v0idpwn, line 89: Without my change, the problematic map decodes to `%{nil => nil}` instead of `%{0 => 0}`, which is the desired behaviour.

lib/protobuf/decoder.ex
```diff
@@ -234,7 +234,15 @@ defmodule Protobuf.Decoder do
...
     embedded_msg = decode(bin, type)
-    val = if map?, do: %{embedded_msg.key => embedded_msg.value}, else: embedded_msg
+
+    val =
+      if map?,
+        do: %{
+          (embedded_msg.key || map_default(prop, :key)) =>
+            embedded_msg.value || map_default(prop, :value)
+        },
+        else: embedded_msg
```
whatyouhide, line 244: Nit but we'll probably want a `do`/`end` here rather than the inline `do:`.

Also, is this gonna be fine with booleans?

PR #393, opened 2025-01-04
Conformance: accept nulls for repeated and maps in JSON decoder

v0idpwn:


v0idpwn:
Fixes the last 2 failing required tests in the conformance suite from protobuf main.

PR #394, opened 2025-01-10
Add native Elixir `JSON` support

v0idpwn:
This implementation defaults to `JSON` if it's available. If it's not,
it tries to use `Jason` as it did before.

Some exceptions are changed (only for `JSON`, `Jason` is unaffected). For
example, instead of `Jason.DecodeError`, when using `JSON` we raise
`Protobuf.JSON.DecodeError`, and encode errors have a more varied shape
(since they can be mostly anything coming from Elixir side).

Additionally, it adds 1.18.1 in CI. Once it's working in both versions, I'll 
remove 1.17.x.

### DIFF
--- a/lib/protobuf/json.ex
+++ b/lib/protobuf/json.ex
@@ -314,11 +314,8 @@ defmodule Protobuf.JSON do
       {:ok, json_data} ->
         from_decoded(json_data, module)
 
-      {:error, exception} when is_exception(exception) ->
+      {:error, exception} ->
         {:error, exception}
-
-      {:error, other} ->
-        {:error, DecodeError.new(other)}
     end
   end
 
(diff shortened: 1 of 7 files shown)


v0idpwn:
If we don't want to default to it, we can make it configurable and keep the default as Jason. But I don't see good reasons for that.

lib/protobuf/json.ex
```diff
@@ -273,8 +269,11 @@ defmodule Protobuf.JSON do
   @spec decode!(iodata, module) :: struct | no_return
   def decode!(iodata, module) do
     case decode(iodata, module) do
-      {:ok, json} -> json
-      {:error, error} -> raise error
+      {:ok, json} ->
+        json
+
+      {:error, error} ->
+        raise error
```
v0idpwn, line 276:
```suggestion
      {:ok, json} -> json
      {:error, error} -> raise error
```

PR #397, opened 2025-01-14
Fix encode_to_iodata/1 not matching encode/1 behaviour

v0idpwn:
Only encode/1 was applying transform in the outer struct


lib/protobuf/encoder.ex
```diff
@@ -19,6 +21,11 @@ defmodule Protobuf.Encoder do
     |> IO.iodata_to_binary()
   end
 
+  defp do_encode_to_iodata(%mod{} = struct) do
+    struct
+    |> encode_with_message_props(mod.__message_props__())
```
v0idpwn, line 26:
```suggestion
    encode_with_message_props(struct, mod.__message_props__())
```

PR #398, opened 2025-01-15
Create Protobuf.Text, create `Protobuf.field_presence/2`

v0idpwn:
This PR introduces support for encoding Protobuf messages into the [textproto format](https://protobuf.dev/reference/protobuf/textformat-spec/). 

### Motivation  

The primary motivation for this feature is to improve the ability to inspect and visualize Protobuf messages, while still conforming to some standard. The inspect protocol default implementation can be painful to work with large or complex structures. Finding the relevant, non-default fields in a inspected struct is not intuitive and takes an unnecessary amount of time. A textproto representation provides a human-readable alternative that is much closer to the actual data transmitted over the wire (specially when using transformers). 

### This implementation

- **Encoding only**: Decoding was deliberately left out as it is less valuable compared to encoding, and has many caveats.
- **Emphasis on readability**: pretty-printed by default. We do a best effort for breaking lines, and opted for the most readable and elixir-like alternatives in the spec. Max line width ~~and tab size are~~ is configurable.
- **Limitations**:  
  - No support for extensions
  - No support for `Google.Protobuf.Any`.  
  - No type validation is performed. 

### Future Considerations  

I'm open to extending this implementation in the future to include:  
1. Adding support for extensions
2. Adding support for `Google.Protobuf.Any`.  
3. Implementing type validation.


lib/protobuf/text.ex
```diff
@@ -0,0 +1,255 @@
+defmodule Protobuf.Text do
+  @moduledoc """
+  Text encoding of protobufs
+
+  According to https://protobuf.dev/reference/protobuf/textformat-spec/, without
+  extensions or `Google.Protobuf.Any` support.
+
+  Useful for inspecting/debugging protobuf-encoded data.
```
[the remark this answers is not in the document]
whatyouhide: Everywhere else too please 😄

lib/protobuf/text.ex
```diff
@@ -0,0 +1,255 @@
...
+
+  Accepts the following options:
+
+  - `:pad_size` - indentation size, in number of spaces
+  - `:max_line_width` - maximum line width, in columns
+
+  Doesn't perform any validations. If input data is invalid, it produces
+  undecodable output.
+  """
+  def encode(struct, opts \\ []) do
```
whatyouhide, line 49: Spec plz

lib/protobuf/text.ex
```diff
@@ -0,0 +1,255 @@
...
+        if Enum.any?(message_props.oneof, fn {oneof_name, _} -> name == oneof_name end) do
+          case value do
+            {field_name, field_value} ->
+              encode_field({field_name, field_value}, syntax, message_props, acc)
+
+            nil ->
+              []
+
+            _ ->
+              raise "Invalid value for oneof `#{inspect(name)}`: #{inspect(value)}"
```
whatyouhide, line 98:
```suggestion
              raise "invalid value for oneof `#{inspect(name)}`: #{inspect(value)}"
```

whatyouhide (commented):
Should we look at using `Inspect.Algebra` to simplify the line splitting? Not sure what the spec of the format is and I don't have the bandwidth to go read it now, but it seems flexible enough that we could break lines in a few different places?

v0idpwn:
Pushed a new version using `Inspect.Algebra`. It indeed simplified code a lot, thanks for the suggestion. 

Some limitations/changes:
- I wasn't able to have different separators on `container_doc` for single line and multi-line. I'm sure it's achievable if I use the lower level functions, but I think having a `,` after every line isn't that bad (it's optional in the spec)
- I wasn't able to define `:pad_size` as I was in the hand-crafted implementation, so it's a fixed 2. 

I'll address the other comments over the week :)

whatyouhide:
@v0idpwn OK sounds good, lmk when this is gonna be ready for re-review. 🙃

lib/protobuf/text.ex
```diff
@@ -0,0 +1,155 @@
...
+  defp skip_field?(:proto3, "", %FieldProps{oneof: nil}), do: true
+  defp skip_field?(:proto3, false, %FieldProps{oneof: nil}), do: true
+
+  # This is actually new. Should it be ported to Protobuf.Encoder?
+  defp skip_field?(:proto3, value, %FieldProps{type: {:enum, enum_mod}, oneof: nil}) do
+    enum_props = enum_mod.__message_props__()
+    [first_tag | _] = enum_props.ordered_tags
+    %{name_atom: name_atom, fnum: fnum} = enum_props.field_props[first_tag]
+    value == name_atom or value == fnum
+  end
```
v0idpwn, line 150: I tried this on Protobuf.Encoder and it broke conformance tests. Despite the fact that in the docs it seems like this is the expected behaviour there as well (just as we skip `0` for non-optional integers, I'd expect it to skip the default enum value...)

v0idpwn:
I think it's okay to be reviewed, @whatyouhide :)

lib/protobuf/text.ex
v0idpwn: I'll investigate this further

v0idpwn:
Actually, I found that there are conformance tests for text encoding. I'll work on them.

test/protobuf/text_test.exs
```diff
@@ -0,0 +1,150 @@
...
+          %TestMsg.Foo.Bar{a: 5},
+          %TestMsg.Foo.Bar{a: 1, b: String.duplicate("Hello", 15)},
+          %TestMsg.Foo.Bar{a: 7}
+        ]
+      })
+
+    assert result == """
+           a: 1
+           e: {
+             a: 1,
```
v0idpwn, line 71: I'm not very happy with this inconsistency: notice that the "outside" entries don't have `,`, but the nested ones do. Tried to make it match but failed. 

This is fine per the spec.

v0idpwn:
@whatyouhide I think it is. I did a rebase, removed stuff that was already merged (mostly related to CI), and tried to leave commits more optimized for a non-squash merge given the large scope of this PR (sorry about that, won't do it again 💀).

I did the google bump here because I wanted to ensure we had the latest conformance tests for text as well. I can move it to a separate PR if you wish.

whatyouhide:
@v0idpwn can we split up `Protobuf.field_presence/2` into its own PR?

v0idpwn:
Will do. Will split the google bump as well.

PR #403, opened 2025-04-13
Create Protobuf.Presence module

v0idpwn:
This PR is on-top of #398.

Adds a unified presence tracking module. Makes the different encoders use it when deciding whether to skip a field or not.

Adds a public API for checking field presence on `Protobuf.Presence.field_presence/2`. This is similar to the feature present on the official protobuf libraries, but with a different API.

In Ruby, for example, fields that have **explicit presence tracking** generate a `has_{field_name}?` method:
```ruby
message = MyMessage.new()
message.has_foo? #=> false
message.foo = 0
message.has_foo? # => true
```

If another field, say `bar`, doesn't have explicit presence tracking (e.g.: it is a non-optional, non-one_of proto3), it simply doesn't generate the `has_bar?` method, so you can't "presence-check" these fields.

My proposed API is a bit different, and allows presence-checking `:bar`, but sometimes giving
a `:maybe` as a response:
```elixir
message = %Message{}
Protobuf.Presence.field_presence(message, :foo)
#=> :not_present
Protobuf.Presence.field_presence(%Message{foo: 0}, :foo)
#=> :present

Protobuf.Presence.field_presence(%Message{}, :bar)
#=> :maybe
Protobuf.Presence.field_presence(%Message{bar: 0}, :bar)
#=> :maybe
Protobuf.Presence.field_presence(%Message{bar: 1}, :bar)
#=> :present
```

## Motivation

I think it makes sense to centralize presence checking for a few reasons:
1. Presence rules become even more complex with protobuf editions and I'd like to add support for them soon, so this refactoring seems opportune
2. Existing logic is a bit ad-hoc and duplicated. Seemed like a good opportunity to improve the code I noticed when adding `Protobuf.Text`.
3. Official protobuf libraries provide APIs for presence checking and I don't see why this one shouldn't

### DIFF
--- a/lib/protobuf/encoder.ex
+++ b/lib/protobuf/encoder.ex
@@ -77,27 +77,14 @@ defmodule Protobuf.Encoder do
           "Got error when encoding #{inspect(struct_mod)}##{prop.name_atom}: #{Exception.format(:error, error)}"
   end
 
-  defp skip_field?(_syntax, [], _prop), do: true
-  defp skip_field?(_syntax, val, _prop) when is_map(val), do: map_size(val) == 0
-  defp skip_field?(:proto2, nil, %FieldProps{optional?: optional?}), do: optional?
-  defp skip_field?(:proto2, value, %FieldProps{default: value, oneof: nil}), do: true
-
-  defp skip_field?(:proto3, val, %FieldProps{proto3_optional?: true}),
-    do: is_nil(val)
-
-  defp skip_field?(:proto3, nil, _prop), do: true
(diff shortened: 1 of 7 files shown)


v0idpwn:
I will merge this to #398, and continue working from there. If you have any feedback, you can give it either here or there :)

whatyouhide:
I love this but I would expose the function on the `Profobuf` module and make `Protobuf.Presence` private, thoughts?

v0idpwn:
Great idea, will do!

PR #405, opened 2025-04-22
Use `ensure_compiled!` instead of `ensure_loaded?` on map typespec macro

v0idpwn:
Instead of raising, holds the file compilation until the module is available.

Solves compilation issue when using `one_file_per_module=true` and map types, where it would raise due to `Code.ensure_loaded?` returning false for a module that wasn't compiled yet.


yordis (approved):

v0idpwn:
Merged the CI changes (with your comments fixed) on #407 and rebased this one.

PR #409, opened 2025-04-24
feat: add full_name callback to proto messages

yordis:


yordis:
It seems that `full_name` could be deterministic at compile time

```proto
syntax = "proto3";
package demo.v1;

message Example {
  message NestedFirstLevel {
    message NestedSecondLevel {}

    NestedSecondLevel nested_second = 1;
  }

  NestedFirstLevel nested_first = 1;
}
```

``go
package demo__test

import (
	"fmt"
	"github.com/stretchr/testify/assert"
	"google.golang.org/protobuf/reflect/protoreflect"
	"testing"
	demov1 "demo/gen/pb/demo/v1"
)

func TestName(t *testing.T) {
	nestedFirstNestedSecond := &demov1.Example_NestedFirstLevel_NestedSecondLevel{}
	nestedFirst := &demov1.Example_NestedFirstLevel{NestedSecond: nestedFirstNestedSecond}
	example := demov1.Example{NestedFirst: nestedFirst}

	exampleName := fullName(&example)
	fmt.Printf("%v", exampleName)

	nestedFirstName := fullName(example.GetNestedFirst())
	nestedFirstNameTop := fullName(nestedFirst)
	fmt.Printf("%v  --- %v", nestedFirstName, nestedFirstNameTop)
	assert.Equal(t, nestedFirstName, nestedFirstNameTop)

	nestedFirstNestedSecondName := fullName(example.GetNestedFirst().GetNestedSecond())
	nestedFirstNestedSecondNameTop := fullName(nestedFirstNestedSecond)
	fmt.Printf("%v --- %v", nestedFirstNestedSecondName, nestedFirstNestedSecondNameTop)
	assert.Equal(t, nestedFirstNestedSecondName, nestedFirstNestedSecondNameTop)
}

func fullName(msg protoreflect.ProtoMessage) protoreflect.FullName {
	return msg.ProtoReflect().Descriptor().FullName()
}
```

**Comment by yordis** [comment]
It seems that Golang does not allow me pass enum values to anypkg, so I think that such full_name isn't in there for the enum values since it always exists inside a message as a type

Screenshot 2025-05-05 at 5 12 40 PM

**Comment by yordis** [comment]
https://protobuf.com/docs/language-spec#fully-qualified-names

File: test/protobuf/protoc/generator/message_test.exs [hunk 1]
```diff
@@ -7,19 +7,19 @@ defmodule Protobuf.Protoc.Generator.MessageTest do
...
 
   test "generate/2 has right name" do
-    ctx = %Context{}
+    ctx = %Context{package: "pkg.name"}
     desc = %Google.Protobuf.DescriptorProto{name: "Foo"}
     {[], [{_mod, msg}]} = Generator.generate(ctx, desc)
-    assert msg =~ "defmodule Foo do\n"
-    assert msg =~ "use Protobuf, protoc_gen_elixir_version: \"#{Util.version()}\"\n"
+    assert msg =~ "defmodule Pkg.Name.Foo do\n"
+    assert msg =~ "use Protobuf, full_name: \"pkg.name.Foo\", protoc_gen_elixir_version: \"#{Util.version()}\"\n"
```
v0idpwn, line 14: Isn't it better to include a function in the template? Since the function must exist anyway (and is currently generated in the use-macro)

yordis: I am unsure, either way is fine by me, I need to make it work first 😭

yordis: About this, I am following the existing code style, being an argument is so that I can enforce the callback inside, similar to the rest of the code

yordis:
As far as I can tell `package` is technically required, a lot of test cases where such information is missing, I wonder if that should be fixed

lib/protobuf.ex
```diff
@@ -175,6 +178,11 @@ defmodule Protobuf do
   """
   @callback transform_module() :: module() | nil
 
+  @doc """
+  Returns the fully qualified name of the descriptor's target.
+  """
+  @callback full_name() :: String.t() | nil
```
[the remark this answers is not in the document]
yordis: it shouldn't let me fix this

yordis:
> ** (RuntimeError) reading from :stdio failed: {:error, {:no_translation, :unicode, :latin1}}

What in the world is that 🤔

sindrip:
Hi @yordis I'm familiar with this error, it is something that came in OTP 26, there is good context here with a fix and a link to the issue in OTP: https://elixirforum.com/t/apparent-regression-reading-binary-data-from-stdio-in-erlang-26/57111/3

If you are simply looking for a way to get CI to pass, this is one way to do it: https://github.com/sindrip/protobuf/pull/2/commits/f99091d9096cd90250d144198fcc4d8ab760963a
```yaml
    env:
      MIX_ENV: test
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      # Set locale to handle encoding issues in newer Elixir versions
      LC_ALL: C.UTF-8
      LANG: C.UTF-8
      ERL_AFLAGS: "-kernel standard_io_encoding latin1"
```

And then you would have to re-generate some files and CI will pass.

Now I have another suggestion, and maybe this is better suited for @whatyouhide or @v0idpwn but there seems to be a way to pass this to the generated escript, as I do need to pass these kernel encoding flags to a lot of generation scripts that I use at work which is tedious.

And that could be done maybe like this: https://github.com/sindrip/protobuf/pull/2/commits/7c4f3f153f6b20d7b3467c7fbfc284f26448d83b
```elixir
  defp escript do
    [
      main_module: Protobuf.Protoc.CLI,
      name: "protoc-gen-elixir",
      emu_args: emu_args()
    ]
  end

  defp emu_args() do
    case :erlang.system_info(:otp_release) do
      v when v in [~c"26", ~c"27"] -> "-kernel standard_io_encoding latin1"
      _ -> ""
    end
  end
```

But it is not thoroughly tested (although it also works in CI.)

yordis:
@v0idpwn @whatyouhide IT IS PASSING!!!!!!!!!! I am so happy right now

georgecoss:
thank you for helping to work toward supporting grpc error details in elixir!

PR #410, opened 2025-04-25
Fix formula for converting Duration into google.protobuf.Duration

v0idpwn:
Can't use `duration.second` because you may have minutes, weeks, etc to consider.

### DIFF
--- a/lib/google/protobuf.ex
+++ b/lib/google/protobuf.ex
@@ -163,15 +163,22 @@ defmodule Google.Protobuf do
 
         iex> from_duration(Duration.new!(second: 1, microsecond: {500_000, 6}))
         %Google.Protobuf.Duration{seconds: 1, nanos: 500_000_000}
+
+        iex> from_duration(Duration.new!(hour: 1, minute: 2, microsecond: {500_000, 6}))
+        %Google.Protobuf.Duration{seconds: 3720, nanos: 500_000_000}
+
+        iex> from_duration(Duration.new!(minute: 2, microsecond: {6_500_000, 6}))
+        %Google.Protobuf.Duration{seconds: 126, nanos: 500_000_000}
     """
     @doc since: "0.15.0"
     @spec from_duration(Duration.t()) :: Google.Protobuf.Duration.t()
     def from_duration(%Duration{} = duration) do
       {microsecond, _precision} = duration.microsecond
+      seconds = div(to_timeout(duration), 1000)
 
       struct(Google.Protobuf.Duration, %{
-        seconds: duration.second,
-        nanos: microsecond * 1_000
+        seconds: seconds,
+        nanos: rem(microsecond, 1_000_000) * 1_000
       })
     end
   end


lib/google/protobuf.ex
```diff
@@ -163,15 +163,22 @@ defmodule Google.Protobuf do
...
+        %Google.Protobuf.Duration{seconds: 3720, nanos: 500_000_000}
+
+        iex> from_duration(Duration.new!(minute: 2, microsecond: {6_500_000, 6}))
+        %Google.Protobuf.Duration{seconds: 126, nanos: 500_000_000}
     """
     @doc since: "0.15.0"
     @spec from_duration(Duration.t()) :: Google.Protobuf.Duration.t()
     def from_duration(%Duration{} = duration) do
       {microsecond, _precision} = duration.microsecond
+      seconds = div(to_timeout(duration), 1000)
```
v0idpwn, line 177: I expected elixir to have a better API for that, but I didn't find one 🤔

lib/google/protobuf.ex
```diff
@@ -163,15 +163,22 @@ defmodule Google.Protobuf do
 
         iex> from_duration(Duration.new!(second: 1, microsecond: {500_000, 6}))
         %Google.Protobuf.Duration{seconds: 1, nanos: 500_000_000}
+
+        iex> from_duration(Duration.new!(hour: 1, minute: 2, microsecond: {500_000, 6}))
+        %Google.Protobuf.Duration{seconds: 3720, nanos: 500_000_000}
+
+        iex> from_duration(Duration.new!(minute: 2, microsecond: {6_500_000, 6}))
+        %Google.Protobuf.Duration{seconds: 126, nanos: 500_000_000}
```
v0idpwn, line 171: These tests would fail on the previous impl

lib/google/protobuf.ex
yordis, line 177: TIL

yordis (approved):

lib/google/protobuf.ex
yordis: yeah... I was waiting for your code review before merging my implementation 😭

lib/google/protobuf.ex
whatyouhide: I don't think this is the correct implementation. `duration.microsecond` is of type `Calendar.microsecond/0`, and the second element of the tuple should be used in the calculation?

whatyouhide: From a couple weeks ago: https://github.com/elixir-lang/elixir/commit/dd608c99fc3baae0ecf6f8f32c072542bc65053a

v0idpwn: TIL :)
I thought it was only info about how many digits we could rely on.

yordis: TIL ... I needed that extra context around precision as well, I was YOLO it

v0idpwn: Okay, @wojtekmach, per our discussion, this is actually correct, right?

v0idpwn: Ok, @whatyouhide, per our discussion, this is actually correct, right?

whatyouhide: > I expected elixir to have a better API for that, but I didn't find one 🤔

@v0idpwn you can use `System.convert_time_unit/3`.

v0idpwn: I didn't mean the `div/2` but the `to_timeout`. I would like to have `Duration.to_unit(duration, :second)`

PR #411, opened 2025-05-01
Remove deprecated `new` functions

v0idpwn:
These functions have been deprecated since Protobuf v0.12


v0idpwn:
I think it could make sense to reintroduce `new/0,1` in a plugin (like the gRPC one) in the future. I know some users would be interested in that. I played with the concept, but couldn't get an API I really liked, so I'm not moving this initiative forward now.

whatyouhide:
@v0idpwn I would wait til the Elixir type system, which would catch quite a few things that `new/0,1` were there for 🙃

whatyouhide:
Thank you @v0idpwn 💟

PR #418, opened 2025-09-26
Fix old TODO's related to minimum elixir version 1.11

sindrip:
There are a few TODO's in the codebase regarding elixir 1.11, as the project has a minimum elixir version of 1.12 we can remove them.


v0idpwn:
Looks great, thank you! 💜

Issue #360, opened 2023-10-31
Any support unusable


hassox:
Referring to: https://github.com/elixir-protobuf/protobuf/pull/329

Just wondering why the pack and unpack on the any protos was removed. https://github.com/elixir-protobuf/protobuf/pull/329 This change makes it very difficult to work with the any protos. There’s no longer any access to the fully qualified name from the generated code. You can’t access the package or the nesting to construct the type url. 

Using the module name is insufficient since you can’t tell from the module name what is part of the package (lower case) and which are nested messages (CamelCase).

This change seems very strange. Every language I’ve ever used that supports any has a pack and unpack function. Any insight on why this was removed without providing a way to access the information to work with these from the generated code?

yordis:
@whatyouhide any thoughts on this one? I am trying to use the `Any` pkg, and I am getting nowhere, I tried multiple things but nothing worked:

```proto
import "google/protobuf/any.proto";

message EventStoreEvent {
  google.protobuf.Any data = 7;
}
```

```elixir
  event = %SomeProtoMessage{}

  payload =
      Protobuf.JSON.encode!(%EventStoreEvent{
        data: %Google.Protobuf.Any{
          type_url: nil, # how to get this value?
          value: nil # what do I suppose to put here?
        },
      })
```

PR #429, opened 2026-01-21
feat: add unpack/2 to Protobuf.Any

yordis:
Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

Related to #360 and #409 this is the safest way I found to introduce the feature without making such big assumptions that it wouldn't work.

I have done this professionally for a while using https://hex.pm/packages/trogon_typeprovider

```elixir
defmodule Umbrella.Grpc.TypeProvider do
  use Trogon.TypeProvider, prefix: "type.googleapis.com/"

  register_type "google.rpc.BadRequest", Google.Rpc.BadRequest
  register_type "google.rpc.ErrorInfo", Google.Rpc.ErrorInfo
  register_type "google.rpc.Status", Google.Rpc.Status

  # ... add more
end
```

Intentionally, I did not make any assumptions around the `type_url` as far as I can tell, it is not required to be `type.googleapis.com/` prefixed.

- What about global type provider?

We could make the type provider module optional and fallback to some `Application.compiled_env!` or something like that, personally, I rather introduce the feature avoiding global assumptions. We can always add a global function that make such assumption instead.

### DIFF
--- a/lib/protobuf/any.ex
+++ b/lib/protobuf/any.ex
@@ -1,5 +1,4 @@
 defmodule Protobuf.Any do
-
   @moduledoc """
   Provides functions for working with the `google.protobuf.Any` type.
   """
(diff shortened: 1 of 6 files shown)


yordis:
@whatyouhide here is the follow up work, I did the most safe way I found that would truely work across different projects, which I also tested there.

I did not make assumptions about the prefix because as far as I can tell, we can't, but, correct me here.

lib/protobuf/any.ex
```diff
@@ -28,6 +27,38 @@ defmodule Protobuf.Any do
...
+        value: <<...>>
+      }
+      Protobuf.Any.unpack(any, MyApp.AnyTypeProvider)
+      #=> {:ok, %MyApp.Events.UserCreated{...}}
+  """
+  @spec unpack(Google.Protobuf.Any.t(), module()) ::
+          {:ok, struct()} | {:error, reason :: any()}
+  def unpack(%Google.Protobuf.Any{type_url: type_url, value: value}, type_provider) do
+    with {:ok, module} <- type_provider.to_module(type_url) do
+      {:ok, module.decode(value)}
```
whatyouhide: `decode/1` can raise, right? So this can also raise. Should we catch that somehow so that the `{:error, _}` spec is more true to the function?

lib/protobuf/any.ex
```diff
@@ -28,6 +27,38 @@ defmodule Protobuf.Any do
...
+        type_url: "type.googleapis.com/myapp.events.UserCreated",
+        value: <<...>>
+      }
+      Protobuf.Any.unpack(any, MyApp.AnyTypeProvider)
+      #=> {:ok, %MyApp.Events.UserCreated{...}}
+  """
+  @spec unpack(Google.Protobuf.Any.t(), module()) ::
+          {:ok, struct()} | {:error, reason :: any()}
+  def unpack(%Google.Protobuf.Any{type_url: type_url, value: value}, type_provider) do
+    with {:ok, module} <- type_provider.to_module(type_url) do
```
whatyouhide: We should validate that this returns at *at least* an atom, I think, to give a nice error message.

yordis:
@whatyouhide done

yordis:
@whatyouhide any opportunity to code review this one? 🙏🏻

